PR #1, opened 2017-01-29
Rewrite

bmwalters:
* General cleanup
* Drops dependency on jQuery (previously provided by steam page scripts)
* Maintains/improves cross-browser compat
* Converts to a more Promise-centric architecture
* Adds .eslintrc
* ~100 fewer LOC


bmwalters:
I tested on Firefox (developer edition) and Chromium. When packaging for Firefox, I would suggest looking into [web-ext](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext). You may also need to allow the Firefox extension through the CSGOFloat-API CORS filter.

float.js
```diff
@@ -1,327 +1,207 @@
...
+        return Promise.resolve(steamListingInfo[listingId]);
     }
+
+    let script = document.createElement('script');
+    script.innerText = `
+        window.postMessage({
+            listingInfo: g_rgListingInfo
+        }, '*');
+    `;
+    document.head.appendChild(script);
```
bmwalters: This is somewhat sub-optimal. I looked into doing it with a background script and `executeScript`, but ran into sandboxing issues there as well. At least this code is only ran the first time you request listing info per page.

icons/icon.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0"?>
+<svg viewBox="0 0 306 306" style="background-color: #4a4a4a" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
+  <path fill="#ffffff" transform="translate(64, 32)" d="m62.371094,79.371094l29.880005,0q7.369995,-31.25 16.869995,-46.5q9.5,-15.25 24.75,-24.060001q15.25,-8.810001 28.75,-8.810001q7.630005,0 11.5,3q3.880005,3 3.880005,7.120003q0,3.75 -2.630005,6.25q-2.619995,2.5 -6.869995,2.5q-4.380005,0 -16.5,-2.619999q-6,-1.25 -9.380005,-1.25q-8.25,0 -14.869995,10.18q-6.630005,10.189999 -10.630005,27.569996l-6.25,26.620003l33,0l-2,8.879997l-33.369995,0l-19.75,79.750008q-9.130005,36.369995 -29.190002,55.119995q-20.059998,18.75 -43.559998,18.75q-7.880005,0 -11.940002,-3.059998q-4.059998,-3.059998 -4.059998,-7.559998q0,-3.880005 2.809998,-6.190002q2.809998,-2.309998 8.059998,-2.309998q2.630005,0 7.75,0.619995q8.75,1.130005 16.75,1.130005q12.380005,0 18,-5.820007q5.630005,-5.809998 11.25,-28.429993l25.380005,-102.000008l-29.880005,0l2.25,-8.879997z"/>
+</svg>
```
bmwalters: So I added an SVG version of the icon so we could use it in the future. The problem is, Firefox doesn't document its support for SVG extension icons, and you have to do it like this in the manifest:
```json
"icons": {
    "16": "icons/icon.svg",
    "32": "icons/icon.svg",
    "48": "icons/icon.svg",
    "96": "icons/icon.svg",
    "128": "icons/icon.svg",
    "256": "icons/icon.svg"
}
```
which sort of sucks.
Chrome [doesn't support SVG extension icons at all](https://bugs.chromium.org/p/chromium/issues/detail?id=29683).

float.js
bmwalters: I just realized we could probably use a `postMessage` from the extension -> the page and only add a single script tag like you do in the current version.

float.js
```diff
@@ -1,327 +1,207 @@
...
+};
 
-    /*
-        Processes a successful float retrieval for the given itemid
-    */
-    function processSuccess(itemid) {
-        // find the corresponding div
-        var floatdiv = $J("#" + itemid + "_floatdiv");
+let processFloatQueue = function() {
+    if (floatQueue.length === 0) { return setTimeout(processFloatQueue, 500); }
```
Step7750: I think the checking interval should be lower than `500` ms, maybe something like `100` to have a more responsive feel.

float.js
Step7750: That is definitely the better way since we're bombarding the DOM `head` right now with needless repetitious scripts.

float.js
```diff
@@ -1,327 +1,207 @@
...
+let floatData = {};
+let floatTimer;
+let steamListingInfo = {};
+
+// retrieve g_rgListingInfo from page script
+window.addEventListener('message', (e) => {
+    steamListingInfo = e.data.listingInfo;
+});
+
+let retrieveListingInfoFromPage = function(listingId) {
```
Step7750: Why not use `const` instead of `let` (and for the rest of the function declarations)?

float.js
```diff
@@ -1,327 +1,207 @@
...
+    allFloatSpan.innerText = 'Get All Floats';
+    allFloatButton.appendChild(allFloatSpan);
+
+    let githubLink = document.createElement('a');
+    githubLink.style.marginLeft = '10px';
+    githubLink.style.textDecoration = 'underline';
+    githubLink.style.fontFamily = `'Motiva Sans', sans-serif`;
+    githubLink.href = 'https://github.com/Step7750/CSGOFloat';
+    githubLink.innerText = 'Powered by CSGOFloat';
+    parentDiv.appendChild(githubLink);
```
Step7750: I must say, the styling here is really great for the separate div!

icons/icon.svg
Step7750: The SVG logo looks awesome, thank you!

manifest.json
```diff
@@ -1,23 +1,22 @@
...
-              "48": "icons/48.png",
-              "128": "icons/128.png"},
-    "description":  "Dedicated API for fetching the float value and paint seed of CSGO items on the market",
-    "name":         "CSGOFloat Market Checker",
-    "short_name":   "CSGOFloat",
-    "version":      "1.0.2"
-}
\ No newline at end of file
+    "permissions": [
+        "https://api.csgofloat.com:1738/*",
```
Step7750: AFAIK, we don't need to request permission for `api.csgofloat.com` since we only need permission to interact with steamcommunity pages.

Step7750:
Also, I think some really great features like item sorting by float could be implemented, but it is hard to do that without conflicting with other big extensions like SIH. Also, highlighting floats if they're below a certain percentile might be useful since the API returns the min/max float for the skin and floats are distributed uniformly across their given wear range.

This is an awesome step in that direction 👍

float.js
```diff
@@ -1,327 +1,207 @@
...
-    console.log('%c Changelog can be found here: https://github.com/Step7750/CSGOFloat ', 'background: #222; color: #fff;');
-}
+// start the queue processing loop
+processFloatQueue();
 
-// Inject the needed functions into the DOM
-injectToDOM();
\ No newline at end of file
+console.log('%c CSGOFloat Market Checker (v1.1.0) by Step7750 ', 'background: #222; color: #fff;');
+console.log('%c Changelog can be found here: https://github.com/Step7750/CSGOFloat ', 'background: #222; color: #fff;');
```
Step7750: `https://github.com/Step7750/CSGOFloat` -> `https://github.com/Step7750/CSGOFloat-Extension`

Step7750:
I just tried out web-ext and its really awesome! It looks like the CORS filter is configured correctly for Firefox on the API server right now; I don't know if the extension, once its published, will have a different `origin` header though.

The Firefox extension has been in the queue for review for quite some time though. It'd be great to upload v1.1 by the time they get around to the review (currently 39 of 328).

float.js
bmwalters: @Step7750 do you prefer using `const` wherever possible or only in cases like this?

manifest.json
bmwalters: I think it's necessary for Firefox (WebExtensions).
From [the MDN page](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/permissions):
> Host permissions are specified as match patterns, and each pattern identifies a group of URLs for which the extension is requesting extra privileges.
The extra privileges include:
    * XHR access to those origins

Step7750: Huh, that's weird, it works perfectly fine without it on the latest stable version of FF. Anyhow, for future compatibility, it'll stay.

Step7750:
I found a little bug, if you switch the page while it is fetching floats, it will error out with the following trace:

```
Uncaught (in promise) TypeError: Cannot read property 'querySelector' of null
    at processFloatQueue (float.js:63)
    at getFloatData.then.catch (float.js:86)
```

bmwalters:
@Step7750 good catch. In that case, do you think we should abort the request or continue to make the request in case they decide to navigate back to the page?

Step7750:
@bmwalters As I said above, I think we should clear the queue. Otherwise, if they want to process floats on the new page, they won't know why its taking so long since it is still processing the old stuff.

Step7750:
Looks good to me, any last minute changes?

bmwalters:
@Step7750 Nope

Step7750:
I'll push it to the Chrome Store and Firefox Add-ons right now.

PR #4, opened 2017-01-29
Fixes Listing Info Retrieval Handling when Slow

Step7750:
Issue: 
Sometimes the page doesn't reply with listing info fast enough and `retrieveListingInfoFromPage` returns null data in the promise.

Fix: 
Delays the timeout for retrieving to 10ms. In the vast majority of
cases, it will be retrieved in under 10ms. If it still takes longer,
`retrieveListingInfoFromPage` rejects the promise. If the promise is
rejected for a single listing, the listing text is updated (to prevent
an infinite loop if they're on a new page). If the promise is rejected
for 'Get all floats', it tries again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 web-ext-artifacts/
+.idea
--- a/float.js
+++ b/float.js
@@ -2,31 +2,31 @@ let floatQueue = [];
 let floatData = {};
 let floatTimer;
 let steamListingInfo = {};
+let listingInfoPromises = [];
 
 // retrieve g_rgListingInfo from page script
 window.addEventListener('message', (e) => {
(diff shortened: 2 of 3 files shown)


bmwalters:
I'd say a better approach would be to store all the promises we need to resolve somewhere, then when we receive the message from the page in the event handler, resolve them all.

Step7750:
@bmwalters I just finished my commit as I saw yours, haha. I think my approach is pretty clean and just stores the resolve objects. If the id isn't in the listing data on an individual listing, it silently fails and the user must have switched pages. This approach also allows us to get rid of the reject handlers like before, but I guess it is helpful to have them so....

Step7750:
@bmwalters Which approach would you go with?

PR #6, opened 2017-01-31
Eliminates Use of innerHTML

### LINKED ISSUE #5: DOM Node Creation/Modification Security
The extension failed to pass Mozilla's review process, due to the following reason:

1) This add-on is creating DOM nodes from HTML strings containing potentially unsanitized data, by assigning to innerHTML, jQuery.html, or through similar means. Aside from being inefficient, this is a major security risk. For more information, see https://developer.mozilla.org/en/XUL_School/DOM_Building_and_HTML_Insertion . Here are some examples that were discovered: float.js#L53, 90

Once #4 is merged, I'll start work on a fix.

Step7750:
Fixes #5 

Removes the use of innerHTML by creating two separate divs for the float and paint seed and populating them independently (instead of a `<br>`). If we need to construct complex HTML in the future, we can use [JSON Templating ](https://developer.mozilla.org/en-US/Add-ons/Overlay_Extensions/XUL_School/DOM_Building_and_HTML_Insertion) (for the filtering, it might be needed).

### DIFF
--- a/float.js
+++ b/float.js
@@ -45,22 +45,24 @@ const getFloatData = function(listingId, inspectLink) {
 const showFloat = function(listingId) {
     let itemInfo = floatData[listingId];
 
-    if (itemInfo) setFloatText(listingId, `Float: ${itemInfo.floatvalue}<br>Paint Seed: ${itemInfo.paintseed}`, true);
-};
-
-const setFloatText = function(listingId, text, removeButton) {
     let floatDiv = document.querySelector(`#item_${listingId}_floatdiv`);
 
     if (floatDiv) {
-        if (removeButton) {
-            // Remove the "get float" button
-            let floatButton = floatDiv.querySelector('.floatbutton');
-            if (floatButton) { floatDiv.removeChild(floatButton); }
-        }
+        // Remove the "get float" button
+        let floatButton = floatDiv.querySelector('.floatbutton');
+        if (floatButton) floatDiv.removeChild(floatButton);
 
-        // Show the text to the user
+        // Remove message div
         let msgdiv = floatDiv.querySelector('.floatmessage');
-        msgdiv.innerHTML = text;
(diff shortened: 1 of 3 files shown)


float.js
```diff
@@ -199,9 +201,14 @@ const addButtons = function() {
...
         getFloatButton.appendChild(buttonText);
 
-        let messageDiv = document.createElement('div');
-        messageDiv.classList.add('floatmessage');
-        buttonDiv.appendChild(messageDiv);
+        // Create divs the following class names and append them to the button div
+        let divClassNames = ['floatmessage', 'itemfloat', 'itemseed'];
+
+        for (let className of divClassNames) {
+            let div = document.createElement('div');
```
bmwalters: I would say semantically a `p` tag makes more sense here

bmwalters:
@Step7750 I believe that page about JSON templating mentions it is deprecated technology. Besides, it probably wouldn't work in Chrome.

Another idea is shipping HTML files with the extension as assets, then use [HTML imports](https://www.html5rocks.com/en/tutorials/webcomponents/imports/) from our script to bring them into the page.

float.js
Step7750: Sadly the `p` tag has styling for padding to the top and bottom of the element. We'd have to customize styling for it either by using JS or injecting a stylesheet. Semantically it's better, but I don't think it is worth the additional manual styling needed.

Just to make sure I understand which tags you want to change, I'm referring to this:
```
<div id="item_284126190785063316_floatdiv" style="display: inline; text-align: left;">
    <p class="itemfloat">Float: 0.2196776419878006</p>
    <p class="itemseed">Paint Seed: 789</p>
</div>
```

Step7750:
@bmwalters The static template strings are much more readable and should be fine security wise. We should definitely use them when possible for the filtering options. Sucks we can't use HTML Imports :(

PR #8, opened 2017-03-28
Fixes Firefox Extension 'Unknown Error'

### LINKED ISSUE #7: `Unknown Error` on Firefox Extension
Bug: Trying to the fetch the float for any item gives `Unknown Error`

Looks like a permissions issue with the request on new versions of Firefox. The request never gets initiated so it can't be a reply origin issue. Firefox doesn't seem to have very descriptive errors for AJAX requests, so this might be a pain.

Step7750:
Fixes #7 

Due to the latest version of Firefox blocking API requests, the fetching is now done in a background script.

### DIFF
--- a/background.js
+++ b/background.js
@@ -1,0 +1,9 @@
+browser.runtime.onMessage.addListener((request, sender, sendResponse) => {
+    fetch(`https://api.csgofloat.com:1738/?url=${request.inspectLink}`)
+    .then((response) => {
+        response.json().then((data) => sendResponse(data));
+    })
+    .catch(() => sendResponse());
+
+    return true;
+});
--- a/background.js
+++ b/background.js
@@ -1,6 +1,4 @@
-let thisBrowser = (window.browser) ? browser: chrome;
-
-thisBrowser.runtime.onMessage.addListener((request, sender, sendResponse) => {
+chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
     fetch(`https://api.csgofloat.com:1738/?url=${request.inspectLink}`)
     .then((response) => {
         response.json().then((data) => sendResponse(data));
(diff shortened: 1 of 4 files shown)


float.js
```diff
@@ -35,10 +35,13 @@ const getFloatData = function(listingId, inspectLink) {
...
     }
 
-    return fetch(`https://api.csgofloat.com:1738/?url=${inspectLink}`)
-    .then((response) => {
-        if (response.ok) return response.json();
-        return response.json().then((err) => { throw err; });
+    return new Promise((resolve, reject) => {
+        let thisBrowser = (window.browser) ? browser: chrome;
+
+        thisBrowser.runtime.sendMessage({'inspectLink': inspectLink}, (data) => {
```
bmwalters: Sould use `chrome.runtime.sendMessage` here in order to use the callback-style API.
Reference:
> "As a porting aid, the Firefox implementation of WebExtensions supports chrome and callbacks as well as browser and promises. This means that many Chrome extensions will just work in Firefox without any changes. However, this is not part of the WebExtensions standard, and might not be supported by all compliant browsers."
> https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Chrome_incompatibilities#Callbacks_and_the_chrome.*_namespace

background.js
```diff
@@ -0,0 +1,11 @@
+let thisBrowser = (window.browser) ? browser: chrome;
```
bmwalters: In a Firefox extension window.browser and window.chrome are both undefined. The global object isn't actually `window` inside of a WebExtension. Thus, this expression will always result in `chrome`

Step7750:
4eb7a19 should resolve those issues, thanks! Do you happen to know what exactly changed that caused fetching cross domain to error out? The `fetch` errors are so vague...

PR #11, opened 2018-08-16
Fixes Selector when User Has Items on Sale

z32nissan:
A Fix for community market item listing pages where the user has items for sale.

Steam added a feature where it displays items you are selling directly on the item page, it broke the plugin on items where you have items listed. I've added a more specific selector to fix the issue.


Step7750:
LGTM, I'll merge next week and push updates on Chrome/FF.

Issue #33, opened 2019-03-29
Show Wear on Applied Stickers in Market/Inventory


Step7750:

Step7750:
Last part is a duplicate of https://github.com/csfloat/extension/issues/137

PR #35, opened 2019-04-13
Implements Showing Sticker Wear on Market Stickers

Step7750:
#33

### DIFF
--- a/float.js
+++ b/float.js
@@ -191,13 +191,16 @@ const showFloat = function(listingId) {
         if (seedDiv) seedDiv.innerText = `Paint Seed: ${itemInfo.paintseed}`;
 
         // Set the wear value for each sticker
-        for (const stickerIndex in itemInfo.stickers) {
+        for (let stickerIndex = 0; stickerIndex < itemInfo.stickers.length; stickerIndex++) {
             const sticker = itemInfo.stickers[stickerIndex];
 
             // Check if the sticker div exists
-            const stickerWearDiv = floatDiv.parentNode.querySelector(`#sticker_${stickerIndex}_wear`);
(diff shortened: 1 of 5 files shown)


float.js
```diff
@@ -190,6 +190,17 @@ const showFloat = function(listingId) {
         let seedDiv = floatDiv.querySelector('.itemseed');
         if (seedDiv) seedDiv.innerText = `Paint Seed: ${itemInfo.paintseed}`;
 
+        // Set the wear value for each sticker
+        for (const stickerIndex in itemInfo.stickers) {
```
Perplex: You should follow Google's style guide for javascript. Use a `for-of` or keep an index variable and just increment it.

https://google.github.io/styleguide/jsguide.html#features-for-loops

float.js
```diff
@@ -493,47 +507,49 @@ const addMarketButtons = function() {
...
+        const steamListingData = await retrieveListingInfoFromPage(id);
+        const listingData = steamListingData[id];
+        if (!listingData) return;
+
+        const assetID = listingData.asset.id;
+        const steamListingAssets = await retrieveListingAssets(assetID);
+
+        const asset = steamListingAssets[assetID];
+        const lastDescription =
+            asset.descriptions[asset.descriptions.length - 1];
```
Perplex: This should be one line, looks weird otherwise.

float.js
```diff
@@ -493,47 +507,49 @@ const addMarketButtons = function() {
...
+        if (
+            lastDescription.type === 'html' &&
+            lastDescription.value.includes('sticker')
+        ) {
```
Perplex: This should be one line, looks weird otherwise.

PR #41, opened 2019-04-14
Fixes Duplicate Stickers on Filter Change

Step7750:
* If you remove/change filters, it will add a new sticker div without removing the old one
* This ia kind of brittle, should be more robust in a refactor


Step7750:
That error seems unrelated, file a new bug if it persists.

PR #54, opened 2019-04-22
Shows Float/Paint Seed in Inventory Boxes At A Glance

Step7750:
* Shows item data within the inventory item box
* Automatically fetches data for visible items
* Cancels queued requests if the page for that item is no longer visible


Step7750:
Screen Shot 2019-04-22 at 12 05 37 PM

PR #84, opened 2020-04-29
Fixes Proof Fetching For Some Languages

Step7750:
* Forces the initial timestamp to be fetched in English, preventing potential parsing issues for some languages.

### DIFF
--- a/lib/time_fetcher.js
+++ b/lib/time_fetcher.js
@@ -17,8 +17,8 @@ function historyRowHashcode_CSGOFloat(row) {
 
 function getTimestampFromTrade_CSGOFloat(row) {
     const dateDiv = row.querySelector('.tradehistory_date');
-    const date = dateDiv.innerText.split('\n')[0];
-    const time = dateDiv.innerText.split('\n')[1];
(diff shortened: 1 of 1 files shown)


lib/time_fetcher.js
```diff
@@ -91,13 +89,37 @@ async function hasTradeBeforeTime_CSGOFloat(hashCode, timestamp) {
...
 }
 
-async function fetchListingTime_CSGOFloat(hashCode) {
-    const btn = document.querySelector(`#verify_${hashCode}_csgofloat`);
+async function fetchEnglishRow_CSGOFloat(index) {
+    let queryParams = location.search;
+    if (queryParams === '') {
+        queryParams = '?l=english';
+    } else {
+        queryParams += '&l=english';
```
Perplex, line 97: What if they already have a language in their query? Wouldn't this result in something like ...&l=french&l=english

Step7750: Sure, but the configurable language parameter is not meant for regular folks and I don't think there's a way to do it as a user within Steam.

Perplex:
LGTM

PR #87, opened 2020-07-02
Implement Float Bulk Queries

Step7750:


Step7750:
It seems like there's a couple of downsides to this approach that are bigger than initially thought:
1) These POST requests can't be cached by CloudFlare nor have efficient TTLs for browser caching. It turns out that there are around 1.7 million requests per hour when removing this, which is around double the current traffic.
2) Contention waits in the queue can easily cause the endpoint to be slower than the alternative of multiple requests since you effectively wait for the slowest request.
3) Some types of duped items or older inspect links that are fetched don't efficiently use the FloatDB cache, causing the queue for bot item data fetches to the GC to become large.

Keeping it the current way doesn't seem to have much issue since Cloudflare will multiplex the HTTP 2 requests together into unified TCP connections. Bandwidth waste is probably higher with the bulk change, so I'll close this for now.

PR #100, opened 2022-01-04
Fix Souvenir packages throwing constant errors

PenguiniVogel:
Fix Souvenir packages throwing constant errors ( TypeError: Cannot read properties of undefined (reading 'length') ) 
market_actions seem to be unavailable for these, so it throws the constant error that it cannot read the length of undefined.

### DIFF
--- a/float.js
+++ b/float.js
@@ -149,7 +149,7 @@ const getAllFloats = function() {
 
             let listingData = steamListingData[id];
 
-            if (listingData.asset.market_actions == null || listingData.asset.market_actions.length === 0) {
+            if (listingData.asset.market_actions === null || listingData.asset.market_actions.length === 0) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


float.js
```diff
@@ -149,7 +149,7 @@ const getAllFloats = function() {
 
             let listingData = steamListingData[id];
 
-            if (listingData.asset.market_actions.length === 0) {
+            if (listingData.asset.market_actions === null || listingData.asset.market_actions.length === 0) {
```
Step7750, line 152: Should be able to use the optional chaining operator here.

```
if (!listingData.asset.market_actions?.length) { ... }
```

Step7750 (changes requested):

Step7750:
Thanks for putting this together!

Step7750 (approved):

PR #141, opened 2023-05-26
fix: prevent loading floats on agents

ayoung19:
image

image

image

### DIFF
--- a/src/lib/components/market/item_row_wrapper.ts
+++ b/src/lib/components/market/item_row_wrapper.ts
@@ -1,4 +1,4 @@
-import {html} from 'lit';
+import {html, nothing} from 'lit';
 
 import {state} from 'lit/decorators.js';
 import {CustomElement, InjectAppend, InjectionMode} from '../injectors';
@@ -8,7 +8,7 @@ import {Asset, ListingData} from '../../types/steam';
 import {gFloatFetcher} from '../../services/float_fetcher';
 import {ItemInfo} from '../../bridge/handlers/fetch_inspect_info';
 import {getMarketInspectLink, inlineEasyInspect, inlineStickers} from './helpers';
-import {formatSeed, renderClickableRank} from '../../utils/skin';
+import {formatSeed, isSkin, renderClickableRank} from '../../utils/skin';
 import {gFilterService} from '../../services/filter';
 import {AppId, ContextId, Currency} from '../../types/steam_constants';
 import {defined} from '../../utils/checkers';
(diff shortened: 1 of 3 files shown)


src/lib/utils/skin.ts
```diff
@@ -108,3 +108,9 @@ export function isSkin(asset: Asset): boolean {
         (a) => a.category === 'Weapon' || (a.category === 'Type' && a.internal_name === 'Type_Hands')
     );
 }
+
+export function isMarketSkin(asset: Asset) {
```
Step7750, line 112: I think we can merge this with `isSkin`. Make the implementation of `isSkin` a OR between the tag parsing and this.

src/lib/components/market/item_row_wrapper.ts
```diff
@@ -137,6 +137,10 @@ export class ItemRowWrapper extends FloatElement {
         }
 
         if (this.itemInfo) {
+            if (this.asset && !isMarketSkin(this.asset)) {
```
Step7750, line 140: We can move this `if` to outside this block (after the inspect link check above)

Step7750 (changes requested):
🎉🎉 Thanks for your first PR! 🎉🎉

Added a few comments, otherwise LGTM.

src/lib/utils/skin.ts
ayoung19: i don't like that because i lowkey don't want to subject the inventory to this hacky check. perhaps i only fall to this if `asset.tags` is undefined?

ayoung19: so it'd look like
```
if (asset.tags) {
   // inventory check
} else {
  // market check
}
```

ayoung19: honestly it does the same thing but it short circuits if tags are defined

src/lib/components/market/item_row_wrapper.ts
ayoung19: my thought process was we probably want to display the error/loading first right?

ayoung19: otherwise i'd be happy to move it

src/lib/utils/skin.ts
Step7750: Yeah, that sounds good to me.

src/lib/components/market/item_row_wrapper.ts
Step7750: Nah, I think we should treat it the same as an item with no inspect link.

ayoung19: oh right i forgot, was tunnel visioned into causing no regressions

Step7750 (approved):

PR #154, opened 2023-07-12
Add fade percentage to supported skins

ayoung19:
added sorting by fade in market

added fade % for all fade skins and fade border for 100% fade (overwrites stattrak/souvenir but idt players would mind):

image

added black border for fade % in high float skins:

image

use math floor instead of round:
image


src/lib/utils/skin.ts
```diff
@@ -110,3 +111,19 @@ export function isSkin(asset: rgAsset): boolean {
               asset.market_hash_name.includes(keyword)
           );
 }
+
+export function getFadePercentage(asset: rgAsset, itemInfo: ItemInfo): number | undefined {
+    const FADE_TYPE_TO_CALCULATOR = {
```
ayoung19, line 116: nit: move this outside of function

src/lib/components/market/item_row_wrapper.ts
```diff
@@ -140,11 +141,17 @@ export class ItemRowWrapper extends FloatElement {
...
+
         if (this.itemInfo) {
             return html`
                 <div>
                     Float: ${this.itemInfo.floatvalue.toFixed(14)} ${renderClickableRank(this.itemInfo)}<br />
-                    Paint Seed: ${formatSeed(this.itemInfo)}
+                    Paint Seed:
+                    ${formatSeed(this.itemInfo)}${fadePercentage !== undefined
+                        ? html`<br />
+                              Fade: ${round(fadePercentage, 5)}%`
```
ayoung19, line 153: nit: dedupe `%`

package-lock.json
Step7750, line 1: I don't think we should be deleting package-lock here :)

src/lib/components/common/item_holder_metadata.ts
```diff
@@ -58,10 +65,17 @@ export abstract class ItemHolderMetadata extends FloatElement {
...
+        const fadePercentage = this.asset && getFadePercentage(this.asset, this.itemInfo);
+
         return html`
             <span>
                 <span class="float">${formatFloatWithRank(this.itemInfo, 6)}</span>
-                <span class="seed">${formatSeed(this.itemInfo)}</span>
+                <span class="seed"
+                    >${formatSeed(this.itemInfo)}
+                    ${fadePercentage !== undefined
+                        ? html`<span class="fade">(${round(fadePercentage, 1)}%)</span>`
```
Step7750, line 76: This is nit-picky, but CS people care a lot about "100%" fades. Rounding 99.95% to 100% isn't considered a "true" fade and will probably lead to some people getting angry :)

Think we should floor to 1 decimal instead to 99.95% becomes 99.9%.

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -90,10 +91,13 @@ export class SelectedItemInfo extends FloatElement {
             return html``;
         }
 
+        const fadePercentage = this.asset && getFadePercentage(this.asset.description, this.itemInfo);
+
         return html`
             <div class="container">
                 <div>Float: ${this.itemInfo.floatvalue.toFixed(14)} ${renderClickableRank(this.itemInfo)}</div>
                 <div>Paint Seed: ${formatSeed(this.itemInfo)}</div>
+                ${fadePercentage !== undefined ? html`<div>Fade: ${round(fadePercentage, 5)}%</div>` : nothing}
```
Step7750, line 100: Same thing here for `floor`

src/lib/components/market/item_row_wrapper.ts
```diff
@@ -140,11 +141,17 @@ export class ItemRowWrapper extends FloatElement {
             return nothing;
         }
 
+        const fadePercentage = this.asset && this.itemInfo && getFadePercentage(this.asset, this.itemInfo);
```
Step7750, line 144: nit: You can move this `fadePercentage` defn to within the `if-block` below which prevents the additional `this.iteminfo` presence check.

src/lib/utils/skin.ts
```diff
@@ -110,3 +111,19 @@ export function isSkin(asset: rgAsset): boolean {
...
+        Fade: FadeCalculator,
```
Step7750, line 117: Let's also quote `Fade` for consistency.

src/lib/utils/skin.ts
Step7750: Honestly, I'm not opposed to keeping this definition inside here since it'll only ever by used by this function.

Step7750 (changes requested):
👍

src/lib/components/common/item_holder_metadata.ts
ayoung19: got it, i wasn’t aware of this

src/lib/utils/skin.ts
ayoung19: i think the auto formatter doesn’t like it

src/lib/utils/skin.ts
ayoung19: got it, a repetitive 3 field object declaration never hurt anyone 🤣

Step7750 (changes requested):

src/lib/components/market/sort_floats.ts
```diff
@@ -33,54 +43,80 @@ export class SortFloats extends FloatElement {
 
     protected render(): HTMLTemplateResult {
         return html`
-            <csgofloat-steam-button .text="${this.buttonText}" @click="${this.onClick}"></csgofloat-steam-button>
+            <csgofloat-steam-button
+                .text="${this.computeButtonText(SortType.FLOAT)}"
+                @click="${() => this.onClick(SortType.FLOAT)}"
+            ></csgofloat-steam-button>
+            <csgofloat-steam-button
```
Step7750, line 50: We shouldn't show this fade button unless the page is for a "fade" item.

Step7750 (changes requested):
Just a small change or two left, almost gtg.

src/lib/components/common/item_holder_metadata.ts
```diff
@@ -58,10 +69,25 @@ export abstract class ItemHolderMetadata extends FloatElement {
...
+        const rank = getLowestRank(this.itemInfo);
+
         return html`
             <span>
                 <span class="float">${formatFloatWithRank(this.itemInfo, 6)}</span>
-                <span class="seed">${formatSeed(this.itemInfo)}</span>
+                <span class="seed"
+                    >${formatSeed(this.itemInfo)}
+                    ${fadePercentage !== undefined
+                        ? html`<span class="fade ${rank && rank <= 5 ? 'csgofloat-shine-fade-text' : ''}"
```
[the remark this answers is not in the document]
Step7750: Error on my part, this is for making the border black for visibility when there's a float rank

src/lib/utils/skin.ts
```diff
@@ -122,12 +124,22 @@ export function getFadePercentage(asset: rgAsset, itemInfo: ItemInfo): number |
...
+                return [calculator, supportedWeapon.toString()];
             }
         }
     }
 }
 
+export function getFadePercentage(asset: rgAsset, itemInfo: ItemInfo): number | undefined {
+    const fadeCalculatorAndSupportedWeapon = getFadeCalculatorAndSupportedWeapon(asset);
+
+    if (fadeCalculatorAndSupportedWeapon !== undefined) {
```
Step7750, line 136: nit: prefer to return early

```
if (!fadeCalculatorAndSupportedWeapon) {
    return;
}

// do rest
```

src/lib/components/market/sort_floats.ts
```diff
@@ -1,12 +1,12 @@
 import {FloatElement} from '../custom';
```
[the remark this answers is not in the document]
ayoung19: can do!

Step7750 (approved):

PR #157, opened 2023-07-19
Adds Support for Firefox MV3

### LINKED ISSUE #155: Support MV3 for Firefox
Right now, we can't publish HEAD to Firefox since it requires MV3 and Mozilla decided to put their own spin on it.

Step7750:
Firefox's MV3 is a little special...

* Since FF doesn't support service workers, we instead have a background page that imports the compiled background script (needed to import as a module)
* Since FF doesn't support `browser.runtime.sendMessage` from a page, uses a `PostMessageBus` dance for messages from the page to the background script. The content script uses `postMessage` and `on('message', ...)` to facilitate communication with the page, and effectively becomes a proxy in the request chain.

Can test locally using `npm run start_ff` and `web-ext run` in the `dist/` directory.

Closes #155


src/lib/components/injectors.ts
```diff
@@ -43,6 +43,12 @@ export function CustomElement(): any {
         if (!inPageContext()) {
             return;
         }
+
+        if (customElements.get(target.tag())) {
```
Step7750, line 47: Note: This is put here since FF emits warnings on re-defining a tag while Chrome ignores

ayoung19 (approved):

PR #158, opened 2023-07-19
feat: add options page

ayoung19:


Step7750 (commented):
Page layout looks good!

There's a couple of remaining pieces
* System for fetching/storing settings (with defaults defined)
* Defining settings for each piece, hooking it to the relevant function
* Adding a "easy" settings button in some places of the UI (ie. in the little float bar in the Steam Market)

I think we should create a new WIP branch to merge the settings PRs into until it is ready for deployment. Otherwise, it makes it difficult to create releases on `master`.

src/pages/wow/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
Step7750, line 1: This `wow` directory appears to be unused

ayoung19: yeah it was just to test that we could add more pages in the future, will delete in next commit

ayoung19:
oh i was planning on finishing up the pr on this branch and merging it when it's ready, it's still kind of a draft

Step7750:
Converted to a draft PR in the meantime, please click `Ready for review` when you're ready!

Putting it all in one PR is fine with me.

src/lib/bridge/handlers/open_options_page.ts
```diff
@@ -0,0 +1,19 @@
+import {RequestHandler} from '../types';
+import {RequestType} from './types';
+import {isFirefox} from '../../utils/detect';
+
+class OpenOptionsPageHandler implements RequestHandler<undefined, void> {
```
Step7750, line 5: There are other handlers like `EmptyRequestHandler` that you could use.

I'd propose an `EmptyHandler` implementation in `main.ts` with the following:

```
export class EmptyHandler<Req> implements RequestHandler<Empty, void> {
    constructor(private type: RequestType, private handler: (sender: MessageSender) => Promise<void>) {}

    getType(): RequestType {
        return this.type;
    }

    handleRequest(sender: MessageSender): Promise<void> {
        return this.handler(sender);
    }
}
```

Then you can instantiate this handler instead.

src/lib/bridge/handlers/open_options_page.ts
```diff
@@ -0,0 +1,19 @@
...
+    getType(): RequestType {
+        return RequestType.OPEN_OPTIONS_PAGE;
+    }
+
+    async handleRequest(request: undefined, sender: chrome.runtime.MessageSender): Promise<void> {
+        if (isFirefox()) {
```
Step7750, line 11: Replace with `runtimeNamespace().runtime.openOptionsPage()`

src/lib/components/market/skin_viewer.ts
```diff
@@ -95,34 +104,36 @@ export class SkinViewer extends FloatElement {
...
-                      </div>
-                  `
-                : nothing}
-            <img
-                class="screenshot"
-                ?hidden="${this.showing !== Showing.SCREENSHOT || !this.response?.screenshotLink}"
-                src="${this.response?.screenshotLink}"
-            />
-        `;
+        return this.isEnabled
```
Step7750, line 107: Prefer to return early to lower indentation.

```
if (!this.isEnabled) {
  return nothing;
}

// do rest
```

src/pages/options/settings/MarketSettings.tsx
```diff
@@ -0,0 +1,35 @@
...
+                <Text fz="xs">Configure CSGOFloat Market Checker features in the market page and its listings.</Text>
+            </Flex>
+
+            <Space h="sm" />
+
+            <Flex gap="xl">
+                <Flex direction="column" w={500}>
+                    <Title order={6}>3D/Screenshot Buttons</Title>
+                    <Text fz="xs">
+                        Shows buttons under a listing to quickly view 3d model or screenshot of the specific item.
```
Step7750, line 23: If we define the options that _are_ possible in a schema (or just their own `Setting<name, type, data_type>`), then we can also define the description and auto-generate the entire page here.

Adding new settings would become super-simple, just adding the new defined `Setting` option and it all "hooks" up and allows toggling it here.

Step7750 (changes requested):

src/lib/bridge/handlers/open_options_page.ts
ayoung19: the one returned by firefox is a promise while chrome’s isnt

src/background.ts
```diff
@@ -29,3 +32,10 @@ chrome.runtime.onMessageExternal.addListener((request, sender, sendResponse) =>
     unifiedHandler(request, sender, sendResponse);
     return true;
 });
+
+gStore.get<SettingsType>(StorageKey.SETTINGS).then((settings) => {
```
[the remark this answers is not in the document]
ayoung19: sounds good

src/lib/components/market/skin_viewer.ts
```diff
@@ -74,8 +76,15 @@ export class SkinViewer extends FloatElement {
     @state()
     private showing: Showing = Showing.NONE;
 
+    @state()
+    private isEnabled = false;
+
     async connectedCallback() {
         super.connectedCallback();
+
+        const settings = await Get(SETTINGS);
```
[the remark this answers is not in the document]
ayoung19: i want to do it the other way but keeping the type inference we get for free by reading the object would be tricky. i would have to do something that react-hook-form does and let me use typed strings to do things like `getSetting(“market.3d-screenshot-button”)`

also would there be race condition problems if we are getting/setting the object individually?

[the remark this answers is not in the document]
ayoung19: but also it might not make sense to do that when the react app would ideally fetch the entire settings object as form state

src/pages/options/settings/MarketSettings.tsx
ayoung19: i agree it would be really nice once we’ve fleshed out more settings and define specific data types but it feels like an over optimization right now

ayoung19: would it be fair to add this once we have a third repeating data type for our settings

ayoung19: actually at the very least i’ll component-ize this boolean setting field rn

src/background.ts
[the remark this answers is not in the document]
ayoung19: so actually a feature i would like to keep is the ability to seamlessly add new settings and not break anything: if we add a field to `DEFAULT_SETTINGS` then next time the service worker is loaded it automatically does this get/set thing. if we encapsulate getters/setters then is it okay to be setting every single time we are getting (just in case default settings has extra fields that settings doesnt have)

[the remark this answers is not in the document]
Step7750: We need to think about whether we want to change the defaults in the future. If you're writing defaults like this, then you're also overwriting the state of whether the user has actually ever  toggled the setting themselves.

Instead, if we just _on the fly_, coerce an `undefined` (which signifies it is unset) into the default value, then we can change defaults in the future.

src/lib/bridge/handlers/open_options_page.ts
```diff
@@ -0,0 +1,12 @@
+import {RequestHandler} from '../types';
+import {RequestType} from './types';
+import {isFirefox} from '../../utils/detect';
+import {EmptyRequestHandler} from './main';
+
+export const OpenOptionsPage = new EmptyRequestHandler<void>(RequestType.OPEN_OPTIONS_PAGE, async () => {
+    if (isFirefox()) {
```
Step7750, line 7: Use `runtimeNamespace().runtime.openOptionsPage()`

src/lib/components/market/skin_viewer.ts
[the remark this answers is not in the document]
Step7750: I think you could take a page from how the current storage system is designed/encapsulated. It is able to provide type safety for defined keys.

src/pages/options/settings/MarketSettings.tsx
Step7750: Sure, I think my solution is a bit of over-engineering for this. But yeah, let's prioritize making it simple to add new options fields.

package.json
```diff
@@ -41,20 +43,32 @@
         "fast-json-stable-stringify": "^2.1.0",
         "file-loader": "^6.2.0",
         "filtrex": "^3.0.0",
+        "fs": "^0.0.1-security",
```
Step7750, line 46: This shouldn't be needed?

https://www.npmjs.com/package/fs/v/0.0.1-security

Step7750 (changes requested):
It seems like the PR doesn't build:

`[webpack-cli] HookWebpackError: Maximum call stack size exceeded`

Also looks like the build rule caught a circular dependency.

dduart3:
may I know what happened with this PR? I want to create something similar

dduart3:
May I know if this PR is going to be merged?

dduart3:
@Step7750

ayoung19:
Also one of the build tools that checks for import cycles wasn't happy, I can take a look there really only is like 1% left to get this merged.

dduart3:
> Also one of the build tools that checks for import cycles wasn't happy, I can take a look there really only is like 1% left to get this merged.

I was just checking this branch, it looked good and built correctly, I was looking to create something like this but this is already too advanced to be discarded imo. I would just change some minor details and add a couple options that I have in mind already

ayoung19:
Basically the CI has a circular dependency check that isn't passing because it doesn't seem to like the import cycles that exist in React (which is normal).

But yeah how about I try fix up those two small issues preventing it to be merged and then you can build on top of that.

dduart3:
> Basically the CI has a circular dependency check that isn't passing because it doesn't seem to like the import cycles that exist in React (which is normal).
> 
> But yeah how about I try fix up those two small issues preventing it to be merged and then you can build on top of that.

Sounds really good man, would be happy to help if you need it of course

ayoung19:
Yeah the bigger issue is the link to the options page doesn't work on Firefox for whatever reason...

dduart3:
> Yeah the bigger issue is the link to the options page (from the listing page) doesn't work on Firefox for whatever reason... A fresh pair of eyes may help with that.

Sure thing man, I'll look into it

ayoung19:
@Step7750 Omg... This was the problem the whole time 😭 

https://github.com/csfloat/extension/blob/262a725729a9f55d6c2890d51da914b7f36e5126/src/lib/bridge/client.ts#L8

dduart3:
> @Step7750 Omg... This was the problem the whole time 😭
> 
> https://github.com/csfloat/extension/blob/262a725729a9f55d6c2890d51da914b7f36e5126/src/lib/bridge/client.ts#L8

is it working now? haha

ayoung19:
No, I totally missed that the content script <-> background messaging isn't supported for firefox. @Step7750 any input on how you want this to be fixed?

Step7750:
> No, I totally missed that the content script <-> background messaging isn't supported for firefox. @Step7750 any input on how you want this to be fixed?

Sorry, it has been a long time (hope you're doing well @ayoung19 !)

Content Script/Background Messaging is supported on Firefox using the postMessage fallback bus:

https://github.com/csfloat/extension/blob/262a725729a9f55d6c2890d51da914b7f36e5126/src/lib/bridge/client.ts#L43

ayoung19:
@Step7750 Oh yes you're right I'm not sure why I came to that conclusion then. I took another look, it's because I was calling `ClientSend(OpenOptionsPage, undefined)` instead of `ClientSend(OpenOptionsPage, {})`. It works now!

image

Also I noticed the 3D screenshot buttons are gone now so it seems like that option is deprecated. Do you have a new options you'd like added? Would love to finally close this PR.

PR #251, opened 2024-09-16
Visual feedback about pagination status

SaloEater:
Hi, I couldn't stand using market without understanding the status of pagination request so I made this small visualization

https://github.com/user-attachments/assets/1ca87271-fcf4-49f9-987d-e034bb6647c2

### DIFF
--- a/src/lib/components/market/response_render_results_status.ts
+++ b/src/lib/components/market/response_render_results_status.ts
@@ -1,3 +1,27 @@
+function overrideOnResponseRenderResults(proto: any) {
+    const originalMethod = proto.OnResponseRenderResults;
+
+    proto.OnResponseRenderResults = function(responseClass: any) {
+        const success = (responseClass && responseClass.responseJSON && responseClass.responseJSON.success) || false;
+        if (!success) {
+            paginationBox.showError();
+        } else {
+            paginationBox.hide();
+        }
+
+        originalMethod.call(this, responseClass);
+    };
+}
+
+function overrideGoToPage(proto: any) {
+    const originalMethodGoToPage = proto.GoToPage;
+
+    proto.GoToPage = function(page: number) {
+        paginationBox.showPending();
+        originalMethodGoToPage.call(this, page);
+    };
+}
+
 export function initResponseRenderResultsStatus() {
     const id = setInterval(() => {
         const searchResultsObject = g_oSearchResults;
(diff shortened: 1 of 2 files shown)


src/lib/components/market/response_render_results_status.ts
```diff
@@ -0,0 +1,155 @@
...
+    const originalMethodGoToPage = proto.GoToPage;
+
+    proto.GoToPage = function(page: number) {
+        paginationBox.showPending();
+        originalMethodGoToPage.call(this, page);
+    };
+}
+
+export function initResponseRenderResultsStatus() {
+    const id = setInterval(() => {
```
Step7750, line 26: I think it'd be better to get something like this abstracted like `Observe` in utils.

For instance, `WaitUntilDefined` and make it a bit more generic.

src/lib/components/market/response_render_results_status.ts
```diff
@@ -0,0 +1,155 @@
+function overrideOnResponseRenderResults(proto: any) {
```
Step7750, line 1: These functions can be included as part of the inner impl of the component, specifically when it starts you can handle getting the proto reference and overriding. Since at the end of the day, it is tightly coupled anyways and isn't capable of operating outside of a global reference to that component.

Step7750, line 1: Please define types for `g_oSearchResults` (see steam.d.ts) -- it doesn't have to be comprehensive but is certainly better than using any.

src/lib/components/market/response_render_results_status.ts
```diff
@@ -0,0 +1,155 @@
...
+    const originalMethod = proto.OnResponseRenderResults;
+
+    proto.OnResponseRenderResults = function(responseClass: any) {
+        const success = (responseClass && responseClass.responseJSON && responseClass.responseJSON.success) || false;
+        if (!success) {
```
Step7750, line 6: nit: Anti-pattern to do:

```
if (!var) {
   // this
} else {
   // that
}
```

instead of 

```
if (var) {
   // that
} else {
   // this
}
```

src/lib/components/market/response_render_results_status.ts
```diff
@@ -0,0 +1,155 @@
+function overrideOnResponseRenderResults(proto: any) {
+    const originalMethod = proto.OnResponseRenderResults;
+
+    proto.OnResponseRenderResults = function(responseClass: any) {
+        const success = (responseClass && responseClass.responseJSON && responseClass.responseJSON.success) || false;
```
Step7750, line 5:
```suggestion
        const success = !!responseClass?.responseJSON?.success;
```

src/lib/components/market/response_render_results_status.ts
```diff
@@ -0,0 +1,155 @@
...
+        const proto = Object.getPrototypeOf(searchResultsObject);
+
+        overrideOnResponseRenderResults(proto);
+        overrideGoToPage(proto);
+
+        clearInterval(id);
+    }, 1000);
+}
+
+class PaginationFailedBox extends HTMLElement {
```
Step7750, line 39: `PaginationFailedBox` seems to be a bit of a misleading term since it also showing "pending" status.

How about `PaginationStatusBox`?

src/lib/page_scripts/market_listing.ts
```diff
@@ -1,7 +1,11 @@
...
 import '../components/market/item_row_wrapper';
 import '../components/market/utility_belt';
+import '../components/market/response_render_results_status';
+import {initResponseRenderResultsStatus} from '../components/market/response_render_results_status';
 
 init('src/lib/page_scripts/market_listing.js', main);
 
-async function main() {}
+async function main() {
+    initResponseRenderResultsStatus()
```
Step7750, line 10: As mentioned in another comment, it's preferrable for a decorator of the component to handle it's own injection into the page, so that you don't need to handle creation manually here.

Then you just need to import the file.

src/lib/components/market/response_render_results_status.ts
```diff
@@ -0,0 +1,155 @@
...
+    private container: HTMLDivElement;
+    private progressBar: HTMLDivElement;
+    private messageDiv: HTMLDivElement;
+    private hideTimeout: number | null = null;
+    private timeoutDuration: number = 1;
```
Step7750, line 44: This has no clear units, for example, it seems to represent 5s increments for show pending and seconds for error.

You should put the units in the variable name (ie. `timeoutMs`) and use it consistently.

Step7750 (changes requested):
Thank you for the contribution, I really appreciate it!

I hadn't realized how poor the state of navigating SCM has gotten -- pressing the "next" page after loading seemed to consistency fail for multiple tries without any extensions.

I've seen the extension receive 1 star reviews by users thinking _the extension_ introduced this issue, so yeah, I'm thinking about how to integrate messaging in the status bar that indicates the request to Steam failed.

Step7750:
Closing due to inactivity.

It seems like SCM has become more reliable lately, so the necessity of this PR has gone down.

PR #264, opened 2024-10-29
Added UI patterns

nick-dimeglio:
Added the ability for the item checker to see if it is a charm and if it is, display the Pattern id.


Step7750:
Thank you for the contribution @nick-dimeglio!

I had a couple of changes to make so that this works on inventories as well (https://github.com/csfloat/extension/pull/266). I'll close out your PR as a result, but thank you so much for putting this together.

PR #273, opened 2024-11-17
Fix: Selected Item Info for Commodities

### LINKED ISSUE #271: Show Stall Prices If Non-Skin (ie. Commodity)

GODrums:
## Motivation

By now CSFloat support selling of many different item types. To reflect this in the inventory, we want to display the "List on CSFloat" or "Listed for $..."-buttons for all supported items.
Fixes #271.

## Description

This PR reworks the `selected_item_info`-element according to the current selling capabilities on CSFloat:
- Show "List on CSFloat"-button for Agents, Sticker, Pins, Patches and Cases
- Change the link of the "List on CSFloat"-button to the actual selling page, i.e. `https://csfloat.com/sell`
- Fixed a bug where the "Charm Detachment Pack" got classified as Charm
- Support stall prices for Agents
Screenshot 2024-11-17 235851

Implementation notes:
- since we now support items without inspect links (i.e. cases), we now have to re-render the element for every newly selected item. I utilitzed the `isLoading`-variable for this purpose, but there might be a better approach?
- I decided on using a white-list approach on support item types to stay compatible when new items are added

### DIFF
--- a/src/lib/utils/skin.ts
+++ b/src/lib/utils/skin.ts
@@ -105,7 +105,15 @@ export function renderClickableRank(info: ItemInfo): TemplateResult<1> {
 }
 
 export function isSellableOnCSFloat(asset: rgAsset): boolean {
-    return isSkin(asset) || isCharm(asset) || isAgent(asset) || isSticker(asset) || isPin(asset) || isPatch(asset) || isCase(asset);
+    return (
+        isSkin(asset) ||
+        isCharm(asset) ||
+        isAgent(asset) ||
+        isSticker(asset) ||
+        isPin(asset) ||
+        isPatch(asset) ||
+        isCase(asset)
+    );
 }
 
 export function isSkin(asset: rgAsset): boolean {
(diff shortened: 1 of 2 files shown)


src/lib/utils/skin.ts
```diff
@@ -113,16 +125,39 @@ export function isSkin(asset: rgAsset): boolean {
...
+export function isAgent(asset: rgAsset): boolean {
+    return isAbstractType(asset, 'Agent', 'Type_CustomPlayer');
+}
+
+export function isSticker(asset: rgAsset): boolean {
+    return isAbstractType(asset, 'Sticker', 'CSGO_Tool_Sticker');
+}
+
+export function isPin(asset: rgAsset): boolean {
+    return isAbstractType(asset, 'Pin', 'CSGO_Type_Collectible');
```
Step7750, line 140: I believe CSGO_Type_Collectible may encompass more than just Pins?

For example, badges like a 10 Year Veteran Coin have an inspect link and are also a collectible: https://steamcommunity.com/id/Step7750/inventory#730_2_37136727351

Step7750 (changes requested):
Awesome, thanks @GODrums !

A couple small comments, but looks good 👍

src/lib/utils/skin.ts
GODrums: You are right! Actually, I just noticed that CSFloat doesn't even support selling pins, so we can just remove this again.

Step7750 (approved):

PR #274, opened 2024-11-26
Visualize Item Floats in a Floatbar (Market + Inventory)

GODrums:
## Motivation

Floats can be a pretty confusing concept at first glance. To illustrate the basics of low/high-floats, we want to introduce a float-bar, which visualizes an item's float according to all possible values in a colored bar.
Solves #9.

## Description

This PR contains a new float-bar component:
- New UI-Component: `FloatBar` (LitElement) inspired by the floatbar on the CSFloat website:
  - Can be re-used as HTML-element in any component: `<float-bar></float-bar>`
  - Bar is limited according to provided min/max-float values
  - White marker at the actual item float
  - Opacity of `0.8` to adjust color-spectrum to Steam's "dark theme"
- Adds `FloatBar` to inventory-page
- Adds `FloatBar` to market-page and limits its width in a responsive manner

Possible follow-ups:
- [ ] Tooltip with more "details" such as the min/max floats in text format
- [ ] Actual Float as "hover" above the float-marker

## Screenshots

Floatbar in the inventory:
image
Floatbar on the market:
image
Floatbar on the market (mobile):
image

### DIFF
--- a/src/lib/components/inventory/selected_item_info.ts
+++ b/src/lib/components/inventory/selected_item_info.ts
@@ -133,7 +133,7 @@ export class SelectedItemInfo extends FloatElement {
     }
 
     renderFloatBar(): TemplateResult<1> {
-        if (!this.itemInfo || !(this.itemInfo.floatvalue > 0)) {
+        if (!this.itemInfo || !this.itemInfo.floatvalue) {
             return html``;
         }
 
--- a/src/lib/components/inventory/selected_item_info.ts
+++ b/src/lib/components/inventory/selected_item_info.ts
@@ -52,6 +52,28 @@ export class SelectedItemInfo extends FloatElement {
                 color: #ebebeb;
                 text-decoration: none;
             }
+
+            .market-float-bar-container {
+                position: relative;
+                width: 100%;
+                height: 8px; 
+                margin: 5px 0;
+            }
+
+            .market-float-bar-marker {
+                position: absolute; 
+                background-color: #d9d9d9; 
+                width: 3px;
+                top: -3px;
+                height: 14px;
+                border-radius: 4px;
+            }
+
+            .market-float-bar {
+                display: inline-block; 
+                vertical-align: top; 
+                height: 100%; 
+            }
         `,
     ];
 
@@ -101,6 +123,7 @@ export class SelectedItemInfo extends FloatElement {
         const containerChildren: TemplateResult[] = [];
 
         if (isSkin(this.asset.description) && this.itemInfo) {
+            containerChildren.push(this.renderFloatBar());
             containerChildren.push(
                 html`<div>Float: ${this.itemInfo.floatvalue.toFixed(14)} ${renderClickableRank(this.itemInfo)}</div>`
             );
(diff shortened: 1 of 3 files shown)


src/lib/components/inventory/selected_item_info.ts
```diff
@@ -130,6 +132,17 @@ export class SelectedItemInfo extends FloatElement {
         return html` <div class="container">${containerChildren}</div> `;
     }
 
+    renderFloatBar(): TemplateResult<1> {
+        if (!this.itemInfo) {
```
Step7750, line 136: I think we probably want to also prevent rendering the float bar unless the float is >0.

For example, some vanilla items with stickers applied on them will have a float of 0 from the API. Not a big deal though.

src/lib/components/market/item_row_wrapper.ts
```diff
@@ -158,7 +160,8 @@ export class ItemRowWrapper extends FloatElement {
 
             return html`
                 <div class="float-row-wrapper">
-                    Float: ${this.itemInfo.floatvalue.toFixed(14)} ${renderClickableRank(this.itemInfo)}<br />
+                    ${this.renderFloatBar()} Float: ${this.itemInfo.floatvalue.toFixed(14)}
```
Step7750, line 163: nit: Would be cleaner to wrap the `Float: ... ${renderClickableRank(...)}` portion in a `span`

src/lib/ui/floatbar.ts
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        .market-float-bar {
+            display: inline-block;
+            vertical-align: top;
+            height: 100%;
+            opacity: 0.8;
+        }
+
+        .market-float-bar:not([style*='background-color: transparent']):first-of-type {
```
Step7750, line 34: I don't believe the `:not(...)` selector is needed here since we don't have an instance of setting an element to a transparent background color?

AFAIK, if a certain wear range isn't applicable, it's getting a 0% width below.

src/lib/ui/floatbar.ts
```diff
@@ -0,0 +1,74 @@
...
+        const dynamicWidth = (this.maxFloat - this.minFloat) * 100;
+
+        return html`
+            <div class="market-float-bar-container" style="left: ${left}%; width: ${dynamicWidth.toFixed(2)}%;">
+                <div style="height: 100%; border-radius: 4px; overflow: hidden; font-size: 0;">
+                    ${this.floatConditions.map(
+                        (cond) => html`
+                            <div
+                                class="market-float-bar"
+                                style="width: ${((Math.min(cond.max, this.maxFloat * 100) -
```
Step7750, line 62: nit: IMO, it'd be a good idea to pull a bit of this calculation out to make it more readable. Making getters for `minFloat` like `minFloatPercentage` would help with the `* 100` scattered around.

Step7750 (approved):
Thanks @GODrums, looks great!

Left a few code-style comments, nits are optional. The tooltip with details such as min/max float would be cool, but it'd probably a decent amount of work to make a re-usable tooltip system -- your choice whether you want to implement that in a follow up!

src/lib/ui/floatbar.ts
GODrums: Yes, you are absolutely right. I played around with making it transparent first, but changing the width seemed more consistent.

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -130,6 +132,17 @@ export class SelectedItemInfo extends FloatElement {
         return html` <div class="container">${containerChildren}</div> `;
     }
 
+    renderFloatBar(): TemplateResult<1> {
+        if (!this.itemInfo || !(this.itemInfo.floatvalue > 0)) {
```
Step7750, line 136:
```suggestion
        if (!this.itemInfo || !this.itemInfo.floatvalue) {
```

Step7750 (approved):

PR #278, opened 2024-12-02
CSS-Only Tooltip Component

### LINKED ISSUE #276: Create Reusable Tooltip / Popover System
## Motivation

To cater a userbase with a large discrepancy in knowledge levels, it is fundamental to make sure to provide the extra bit of technical details while not overwhelming casuals. A useful technique is hiding additional information in optional elements, such as tooltips and popovers.

## Description

We want to create a reusable and dynamic tooltip / popover system, which can easily be integrated into other components. 
It should support:
- custom "trigger" elements, for example an icon
- adjustable tooltip container with a base degree of responsiveness

Considered Options:
- [HTML popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API)
- Custom JS logic + positioning library like [floating-ui](https://github.com/floating-ui/floating-ui)

GODrums:
## Motivation
Closes #276.

## Description

This PR introduces a new Tooltip-Component:
- Uses [hint.css](https://github.com/chinchang/hint.css) (unminified, with animations + themes, ~8KB) with custom styling imitating Steam's tooltips
- Inputs for the tooltip class (useful for hint.css's classes) and the rendered label text
- Adds a first sample tooltip for a description of the floatbar:
  - Equivalent to the one on CSFloat including the float ranges
  - Works in inventory + market
  - Inventory: let the selectedItem-container overflow for the tooltip
- Move the Floatbar-component in the `/common/ui`-folder

## Screenshots
Example tooltip from Steam:
image
New Floatbar tooltip using the Tooltip-component (Inventory):
image2
New Floatbar tooltip using the Tooltip-component (Market):
image3


src/lib/components/common/ui/hintcss.ts
```diff
@@ -0,0 +1,392 @@
+import {css} from 'lit';
+export const hintcss = css`
+    /*! Hint.css - v3.0.0 - 2023-11-29
```
Step7750, line 3: We need to include the LICENSE as well. https://github.com/chinchang/hint.css/blob/master/LICENSE

Perhaps we can move this to `src/thirdparty/hintcss/...` and include the .ts + LICENSE file there?

src/lib/components/common/ui/tooltip.ts
```diff
@@ -0,0 +1,47 @@
+import {css, html} from 'lit';
+import {FloatElement} from '../../custom';
+import {CustomElement} from '../../injectors';
+import {property} from 'lit/decorators.js';
+
+import {hintcss} from './hintcss';
+
+@CustomElement()
+export class Tooltip extends FloatElement {
+    @property({type: String}) class = '';
```
Step7750, line 10: This isn't used anywhere -- I'd assume it's just meant to add extra classes to the tooltip?

Perhaps rename to `extraClass` or `extraClasses` with a description of where you can find the available classes to use.

src/lib/components/common/ui/tooltip.ts
```diff
@@ -0,0 +1,47 @@
...
+                }
+            `,
+        ];
+    }
+
+    render() {
+        const tooltipClass = `hint--top hint--rounded hint--no-arrow ${this.class}`;
+
+        return html`
+            <div class="${tooltipClass}" aria-label="${this.text}" style="display: block;">
```
Step7750, line 42: It seems like we could avoid forcing the caller to a somewhat _rigid_ DOM structure that requires a `block` element if we move to using a Lit Directive (https://lit.dev/docs/templates/custom-directives/).

This would potentially change it to something like:

```
<div class="market-float-bar-container" style="left: ${left}%; width: ${dynamicWidth.toFixed(2)}%;">
    ${tooltip("hey this is a tooltip")}
    ...
</div>
```

You'd need to include `hintcss` in the base FloatElement styles I believe. I don't have a strong preference of using a directive however, so if it proves to be annoying then we can ditch it -- just figured we should try to reduce the DOM complexity.

src/lib/components/common/ui/floatbar.ts
```diff
@@ -0,0 +1,103 @@
+import {html, css} from 'lit';
+import {property} from 'lit/decorators.js';
+import {FloatElement} from '../../custom';
+import {CustomElement} from '../../injectors';
+import './tooltip';
+
+@CustomElement()
+export class FloatBar extends FloatElement {
```
Step7750, line 8: Thanks for updating this to extend from FloatElement, forgot to catch that earlier.

Step7750 (changes requested):

src/lib/components/common/ui/tooltip.ts
GODrums: Sorry for the late reply! I absolutely do see your point here. As this is my first project with Lit, I wasn't quite familiar with Directives yet.

I switched to a Lit Directive for the tooltip now in 4eae18c. It indeed does look a bit cleaner, especially in regards to the DOM structure.

So the workflow with the new implementation is now:
- import the directive and use it as `${tooltip("this is a tooltip", "optional-extraclass")}`
- import `tooltipStyles` and add it in the component-styles with `static styles = [...tooltipStyles]`, which contains both `hintcss` as well as the custom tooltip styling

Do let me know if this matches your initial idea with Directives.

src/lib/components/common/ui/tooltip.ts
GODrums: Exactly, `hintcss` provides a few classes to customize the tooltip styling, positioning and functionality.
Since there are too many to fit into a compact comment, I added a reference link for it in 7f1a30c.

src/lib/components/common/ui/tooltip.ts
Step7750: Thanks @GODrums ! Overall, it looks good!

I think you can just put the tooltip styles by default in `FloatElement` and can rename `tooltipDirective` to just `tooltip`.

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -222,5 +226,11 @@ export class SelectedItemInfo extends FloatElement {
                 .fetch({steam_id64: g_ActiveInventory?.m_owner.strSteamId})
                 .then((stall) => (this.stall = stall));
         }
+
+        // Make sure the parent container can overflow
+        const parentContainer = document.querySelector<HTMLElement>('.item_desc_content');
```
Step7750, line 231: We should be finding this element relative to the current element, instead of query selecting across the whole document.

src/lib/components/common/ui/tooltip.ts
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        const newParentClass = `${this.parentNode.getAttribute('class') || ''} hint--top hint--rounded hint--no-arrow ${
+            this.extraClasses
+        }`;
+
+        this.parentNode.setAttribute('class', newParentClass);
+        this.parentNode.setAttribute('aria-label', this.label);
+    }
+    render(label: string, extraClasses?: string) {}
```
Step7750, line 29: nit: new line

src/lib/components/common/ui/floatbar.ts
```diff
@@ -1,43 +1,50 @@
...
-    static styles = css`
-        .market-float-bar-container {
-            position: relative;
-            width: 100%;
-            height: 8px;
-            margin: 5px 0;
-        }
+    static styles = [
+        ...FloatElement.styles,
+        ...tooltipStyles,
```
Step7750, line 15: I'm fine with importing `tooltipStyles` inside of `FloatElement` and also importing `tooltip` in `FloatElement` so any element can just use the directive.

Step7750 (changes requested):

src/lib/components/inventory/selected_item_info.ts
GODrums: I reworked it with `this.closest(...)` now, I assume this is what you had in mind?

Step7750 (approved):
👍

PR #279, opened 2024-12-05
Fix: Steam Market Sticker Container

GODrums:
## Motivation

In today's update, Steam introduced a new inline display for stickers and charms. Since the extension has its own sticker/charm container, we have to solve the conflict and fix all alignments.

## Description

Since the extension's own container has more features, we want to adopt Steam's unique item popover while purging their new container in favor of ours:
- Remove the new "Inspect In Game"-button in favor of our own inspect-link on-hover on the item image
- Remove Steam's new sticker/charm container
- Fix alignment of floatbar and overflow of tooltip according to new structure
- Fix a bug where keychains did not get displayed because of an incorrect exit check
- Add item popover on-hover to the extension's sticker/charm container
image


Step7750 (approved):
Thanks @GODrums !

PR #283, opened 2024-12-13
Display Metadata on `/tradeoffers`-pages

### LINKED ISSUE #143: Show Float Values on Trade Offers Page
Note, this is not the same as the page as when you're creating/responding to a specific trade offer.

We should show floats and such for the item boxes when looking at https://steamcommunity.com/id/me/tradeoffers/

We currently show it when you actually click on the trade offer only.

GODrums:
## Motivation

Closes #143.

## Description

This PR adds the well-known enhancements from tradeoffer-popouts to the `/tradeoffers`-overview:
- `FetchSteamTrades`-handler to fetch the data from the Steam API through the access token
- `trade_offers.ts` page script:
  - Calculates and stores the api data in an accessible format in the DOM
  - Utilizes the `localStorage` as cache to avoid refetching from the API if there are no new trades since the last fetch
- `TradeOfferHolderMetadata` Lit-component, which reads the required properties from the DOM
- Removes the check regarding a CSFloat account for the "offer tracking on csfloat"-button. Since all Steam-API features rely on the same host permission, enabling the "offer tracking" is a mandatory pre-condition

## Screenshots
Received tradeoffer example:
Screenshot 2024-12-13 211319
Sent tradeoffer example:
Screenshot 2024-12-13 211311


src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+
+            if (!instanceId) {
+                continue;
+            }
+
+            const description = steamTrades.descriptions.find(
+                (d) => d.classid === classId && d.instanceid === instanceId
+            );
+            if (description) {
+                tradeItem.setAttribute('data-description', JSON.stringify(description));
```
Step7750, line 70: Let's add a `csfloat` prefix to all data attributes to prevent collisions.

ie. `data-csfloat-description`

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+            let apiItem = trade?.given_asset_ids?.find((a) => a.classid === classId && a.instanceid === instanceId);
+            if (!apiItem) {
+                isOwnItem = false;
+                apiItem = trade?.received_asset_ids?.find((a) => a.classid === classId && a.instanceid === instanceId);
+            }
+            const ownerId = isOwnItem
+                ? JSON.parse(document.getElementById('application_config')?.dataset?.userinfo || '{}').steamid
+                : trade?.other_steam_id64;
+
+            tradeItem.setAttribute('data-owner-steamid', ownerId);
```
Step7750, line 83: `data-csfloat-owner-steamid`

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+            if (apiItem?.assetid) {
+                tradeItem.setAttribute('data-assetid', apiItem.assetid);
```
Step7750, line 85: `data-csfloat-assetid`

src/lib/alarms/trade_offer.ts
```diff
@@ -6,6 +6,24 @@ import {AnnotateOffer} from '../bridge/handlers/annotate_offer';
 import {PingCancelTrade} from '../bridge/handlers/ping_cancel_trade';
 import {CancelTradeOffer} from '../bridge/handlers/cancel_trade_offer';
 import {FetchSteamUser} from '../bridge/handlers/fetch_steam_user';
+import {rgDescription} from '../types/steam';
+import {HasPermissions} from '../bridge/handlers/has_permissions';
+
+export interface ExtendedOfferStatus {
+    offer_id: string;
+    state: TradeOfferState;
+    given_asset_ids?: ExtendedSingleOffer[];
```
Step7750, line 15: `asset_id` isn't the right name for this since it actually is a list of item attributes including the asset ID.

`given_assets` or `given_items` would have better meaning.

src/lib/alarms/trade_offer.ts
```diff
@@ -6,6 +6,24 @@ import {AnnotateOffer} from '../bridge/handlers/annotate_offer';
...
+    received_asset_ids?: ExtendedSingleOffer[];
```
Step7750, line 16: Similar here

src/lib/alarms/trade_offer.ts
```diff
@@ -6,6 +6,24 @@ import {AnnotateOffer} from '../bridge/handlers/annotate_offer';
 import {PingCancelTrade} from '../bridge/handlers/ping_cancel_trade';
 import {CancelTradeOffer} from '../bridge/handlers/cancel_trade_offer';
 import {FetchSteamUser} from '../bridge/handlers/fetch_steam_user';
+import {rgDescription} from '../types/steam';
+import {HasPermissions} from '../bridge/handlers/has_permissions';
+
+export interface ExtendedOfferStatus {
```
Step7750, line 12: Why not use `TradeOffersAPIOffer` instead of a near-duplicate interface? It seems like this parses `accountid_other` into a SteamID64, but that could just be handled by the caller if they want that format.

src/lib/alarms/trade_offer.ts
```diff
@@ -222,6 +248,30 @@ function offerStateMapper(e: TradeOffersAPIOffer): OfferStatus {
     } as OfferStatus;
 }
 
+function extendedOfferStateMapper(e: TradeOffersAPIOffer) {
```
Step7750, line 251: Is there actually a need for this transform? It doesn't seem to provide value beyond processing steamid32 to steamid64 (which we can get the caller do to if they need it)

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+
+    localStorage.setItem('g_steamTrades', JSON.stringify(steamTrades));
+    return steamTrades;
+}
+
+/**
+ * Fetches the api data for trade offers and stores relevant data in the DOM to be used by Lit components.
+ */
+async function getAndStoreTradeOffers() {
+    const isSentPage = location.pathname.includes('sent');
```
Step7750, line 44: I'm trying to think of a good way to avoid having to detect specific pages like this (since URL string matching is somewhat brittle).

For getting the trade offers below, you can just check both `received` and `sent` fields.

However, it seems like we need it to perform the "latest trade offer" check to potentially break the cache. I guess the alternative would be checking if there are any trade offer IDs on the page that don't exist in the API response?

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+
+    const steamTrades = await fetchTradeOffers(isSentPage);
+
+    const tradeOffers = document.querySelectorAll('.tradeoffer');
+
+    for (const tradeOffer of tradeOffers) {
+        const tradeId = tradeOffer.id.split('_')[1];
```
Step7750, line 51: `tradeOfferID`

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+        const tradeItems = tradeOffer.querySelectorAll('.trade_item');
```
Step7750, line 52: `tradeItemElements`

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+        const trade = isSentPage
+            ? steamTrades.sent.find((t) => t.offer_id === tradeId)
+            : steamTrades.received.find((t) => t.offer_id === tradeId);
+
+        for (const tradeItem of tradeItems) {
```
Step7750, line 57: `tradeItemElement`

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+            const economyItemParts = tradeItem.getAttribute('data-economy-item')?.split('/');
```
Step7750, line 58: Could you add a comment showing what the format of this looks like?

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+            const classId = economyItemParts?.[2];
+            const instanceId = economyItemParts?.[3];
+
+            if (!instanceId) {
```
Step7750, line 62:
```suggestion
            if (!classId || !instanceId) {
```

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+            }
+
+            let isOwnItem = true;
+            let apiItem = trade?.given_asset_ids?.find((a) => a.classid === classId && a.instanceid === instanceId);
+            if (!apiItem) {
+                isOwnItem = false;
+                apiItem = trade?.received_asset_ids?.find((a) => a.classid === classId && a.instanceid === instanceId);
+            }
+            const ownerId = isOwnItem
+                ? JSON.parse(document.getElementById('application_config')?.dataset?.userinfo || '{}').steamid
```
Step7750, line 80: Let's move getting the logged in users' Steam ID from the page into a utility function that can be used elsewhere.

We generally parse `g_steamID` from the page instead of this application config, which is IMO less _likely_ to change than the application config (which is a more recent change on steam). I don't have strong opinions on this however.

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+    const steamTrades = await ClientSend(FetchSteamTrades, {});
+
+    localStorage.setItem('g_steamTrades', JSON.stringify(steamTrades));
+    return steamTrades;
+}
+
+/**
+ * Fetches the api data for trade offers and stores relevant data in the DOM to be used by Lit components.
+ */
+async function getAndStoreTradeOffers() {
```
Step7750, line 43: nit: I think `annotateTradeOfferItemElements` may encapsulate what this does a bit better.

Step7750 (changes requested):

src/lib/page_scripts/trade_offers.ts
GODrums: I don't think we have access to the JS-environment of the main page in page scripts. At least it didn't work in my tests:
```
trade_offers.ts:82 Uncaught (in promise) ReferenceError: g_steamID is not defined
```
So for now I moved `application-config`-parsing into its own utility function.

src/lib/page_scripts/trade_offers.ts
GODrums: We can easily remove the URL check in favor of a more costly check of both the `received` and `sent` fields.

I believe checking only the latest trade offer ID should be enough as older ones are guaranteed to be contained in the API response as a consequence. Although I haven't tested how this behavior changes for large amounts of open trades, in which case we would need extra handling for the refetch anyways though.

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,96 @@
...
+ * Gets the trade offers from the local storage or fetches them from the API.
+ * Local storage serves as a cache here.
+ * @param isSentPage if the current page is the sent trade offers page
+ * @returns the trade offers
+ */
+async function fetchTradeOffers(isSentPage: boolean) {
+    const g_steamTrades = JSON.parse(localStorage.getItem('g_steamTrades') || '{}') as FetchSteamTradesResponse;
+    let refetchRequired = true;
+    if (g_steamTrades.sent || g_steamTrades.received) {
+        const latestTradeId = Number.parseInt(g_steamTrades[isSentPage ? 'sent' : 'received']?.[0].offer_id);
```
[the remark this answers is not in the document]
GODrums: I would really like the CachedHandler to be the solution, but I think wouldn't quite do the job for two reasons:
1. sent / received pages display different trade offers, hence the trade_offer_id would be different
2. the cache is not persistent across browser restarts

I gave it a try in 0d4620b, but I couldn't get it to cache the requests correctly even for the same parameters.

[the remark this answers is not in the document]
Step7750: Apologies, it looks like there was a bug in `CachedHandler` that caused this behaviour. I've sent over #284 .

In relation to the other points, for 1), since the trade_offer_id is used within the cache key it should have a different cache between sent/received -- so this shouldn't be an issue.

For 2), this isn't a huge concern since people generally aren't restarting their browsers. In the worst case, we do an extra fetch before 10min.

Let me know if you encounter other issues with this, it seems to work correctly with #284 when tested locally!

src/lib/page_scripts/trade_offers.ts
Step7750: IMO, we can keep the current approach in relation to this. It just might be a _bit_ more brittle, but Valve rarely changes this page (I think it's been at least 8+ years since this has changed).

Let me know if you think otherwise, I'm fine with a separate cache for sent/received tabs since the latest trade offer id is different.

src/lib/page_scripts/trade_offers.ts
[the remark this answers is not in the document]
GODrums: Added additional edge case handling in 48bf2b6 to avoid unnecessary fetches whenever there are no open trades.

Otherwise the new version seems to finally work as intended now!

Step7750 (changes requested):

src/lib/page_scripts/trade_offers.ts
GODrums: If we decide against URL-parsing to unify the cache, our current approach should be the most stable solution in my opinion. 
As I'm not very familiar with Steam's rate limits, I'd trust your experience here. Initially I was a little worried due to other extensions requesting the same data for their functionality. Since we are now storing it in the DOM, it should make reusing it fairly simple though.

src/lib/components/trade_offers/better_tracking.ts
```diff
@@ -53,29 +53,19 @@ export class BetterTrackingWidget extends FloatElement {
...
-            const trades = await ClientSend(FetchPendingTrades, {state: 'queued,pending,verified', limit: 1});
-            if (trades.count === 0) {
-                // They aren't actively using CSFloat Market, no need to show this
-                return;
-            }
-
-            this.show = true;
-        } catch (e) {
-            console.info('user is not logged into CSFloat');
+        if (hasPermissions.granted) {
```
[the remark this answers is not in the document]
GODrums: Fair, I didn't really consider that case!
I reverted it in 42685e6 again.

src/lib/components/trade_offers/trade_offer_holder_metadata.ts
```diff
@@ -0,0 +1,21 @@
...
+@CustomElement()
+// Items in received/sent trade offers
+@InjectAppend('.tradeoffer .trade_item', InjectionMode.CONTINUOUS)
+export class TradeOfferHolderMetadata extends ItemHolderMetadata {
+    get assetId(): string | undefined {
+        return $J(this).parent().attr('data-csfloat-assetid');
+    }
+
+    get asset(): rgAsset | undefined {
+        return JSON.parse($J(this).parent().attr('data-csfloat-description') ?? '{}') as rgAsset;
```
Step7750, line 15: Should probably return `undefined` if the attribute doesn't exist, instead of casting an empty dict to an `rgAsset`

src/lib/page_scripts/trade_offers.ts
```diff
@@ -1,15 +1,105 @@
...
 
 function main() {}
 
+/**
+ * Gets the trade offers from the local storage or fetches them from the API.
+ * Local storage serves as a cache here.
+ * @param steam_id the steam id of logged in user
+ * @returns the trade offers
+ */
+function fetchTradeOffers(steam_id: string) {
```
Step7750, line 21: nit: annotate return type

Step7750 (approved):
LGTM

PR #284, opened 2025-01-07
Fixes Unique Job ID Calc within CachedHandler

Step7750:
Previously it included the sender in the ID, however it seems like browsers such as Chrome have changing properties in this for each request.

Now it only caches based on the request data, tested locally using PR #283


GODrums (approved):
Great find, LGTM!

PR #287, opened 2025-02-01
Fix: Compatibility issues for Firefox

### LINKED ISSUE #243: Support Extension on Firefox
Popup "Setup Trade Verification" says this:
> At the moment, you need to use a Chromium browser (ie. Chrome, Opera GX) in order to use the CSFloat extension. In the future, Firefox will support it -- sorry!

Version 4.3.1 for Firefox is already on GitHub Releases, meaning it *technically* supports Firefox. Why not just submit new version to Mozilla and make it available in Firefox?

GODrums:
## Motivation

Fixes #243.

## Description

With this PR we want to start supporting **Firefox**-based browsers for the new versions of the extension again. To achieve this, this PR changes:
- New **popup** exclusive to the Firefox version to provide an option of requesting the required Steam permissions as the BetterTracking-widget doesn't work in Firefox
- Adjust declarativeNetRequest-rules according to Firefox's random IDs
- Change minimum Firefox version to 127.0 to guarantee permissions listed in `host_permissions` are granted on install (source: https://extensionworkshop.com/documentation/develop/manifest-v3-migration-guide/)
- Chore: update `@types/chrome` to `0.0.300` to successfully use async/await on `chrome.permissions`-methods

Possible TODOs:
- Adjust the BetterTracking "Enable"-button for Firefox to refer to the popup (automatic opening not possible in Firefox)
- Further popup features: 
  - Error state for popup in case of a permission deny
  - More detailed permission status
  - Explanatory text / links / CSFloat socials 

## Screenshots

Popup with missing permissions:
image
Popup with granted permissons:
image

### DIFF
--- a/src/popup/popup.html
+++ b/src/popup/popup.html
@@ -35,6 +35,8 @@
             }
             button:disabled {
                 cursor: not-allowed;
+                background-color: rgba(255, 255, 255, .12);
+                color: rgba(255, 255, 255, .5);
             }
         </style>
     </head>
--- a/src/popup/popup.html
+++ b/src/popup/popup.html
@@ -1,0 +1,29 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <style>
+        body {
+            width: 300px;
+            padding: 16px;
+            font-family: Arial, sans-serif;
+        }
+        button {
+            width: 100%;
+            padding: 10px;
+            background-color: #4CAF50;
+            color: white;
+            border: none;
+            border-radius: 4px;
+            cursor: pointer;
+        }
+        button:hover {
+            background-color: #45a049;
+        }
+    </style>
+</head>
+<body>
+    <button id="requestPermissions">Grant Required Permissions</button>
+    <script src="popup/popup.js"></script>
+</body>
+</html>
--- a/src/popup/popup.html
+++ b/src/popup/popup.html
@@ -59,7 +59,7 @@
                 <path d="M12 8v4" />
                 <path d="M12 16h.01" />
             </svg>
-            <span>Grant Required Permissions</span>
+            <span>Enable Offer Tracking</span>
         </button>
         <script src="popup/popup.js"></script>
     </body>
--- a/src/popup/popup.html
+++ b/src/popup/popup.html
@@ -1,29 +1,64 @@
 <!DOCTYPE html>
 <html>
-<head>
-    <meta charset="UTF-8">
-    <style>
-        body {
-            width: 300px;
-            padding: 16px;
-            font-family: Arial, sans-serif;
-        }
-        button {
-            width: 100%;
-            padding: 10px;
-            background-color: #4CAF50;
-            color: white;
-            border: none;
-            border-radius: 4px;
-            cursor: pointer;
-        }
-        button:hover {
-            background-color: #45a049;
-        }
-    </style>
-</head>
-<body>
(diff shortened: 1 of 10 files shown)


src/popup/popup.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    const hasPermissions = await chrome.permissions.contains({
+        origins: ['*://*.steampowered.com/*'],
+    });
+
+    if (hasPermissions) {
+        // If permissions are already granted, disable the button
+        requestButton.children[1].textContent = 'Permissions already granted';
+        requestButton.setAttribute('disabled', 'true');
```
Step7750, line 14: We should also make the button appear disabled.

The site has disabled buttons that look like this:

image


Background color: rgba(255, 255, 255, .12)
Text Color: rgba(255, 255, 255, .5)

Step7750 (approved):
Thanks @GODrums, Firefox users shall rejoice!

src/popup/popup.html
```diff
@@ -0,0 +1,64 @@
...
+                stroke-linecap="round"
+                stroke-linejoin="round"
+            >
+                <path
+                    d="M20 13c0 5-3.5 7.5-7.66 8.95a1 1 0 0 1-.67-.01C7.5 20.5 4 18 4 13V6a1 1 0 0 1 1-1c2 0 4.5-1.2 6.24-2.72a1.17 1.17 0 0 1 1.52 0C14.51 3.81 17 5 19 5a1 1 0 0 1 1 1z"
+                />
+                <path d="M12 8v4" />
+                <path d="M12 16h.01" />
+            </svg>
+            <span>Grant Required Permissions</span>
```
Step7750, line 60: I think we should make this `Enable Offer Tracking` to match the Chrome Extension's messaging.

Step7750 (changes requested):
Almost forgot, I believe we may need to hide the `enable offer tracking` banner that Chrome gets (or replace it with something that tells them to open the extension popover in the top right).

AFAIK, we cannot programmatically open the extension popover (if we can, that'd be preferrable). I'll add custom instructions for FF users in the front-end.

src/popup/popup.ts
GODrums: Good point, I added it to the CSS in the html-file now.
image

GODrums:
@Step7750 
Those are very good points! I adjusted the button in the following way now:
- Change text to "Enable in the Extension Popup"
- Switch the button to Steam's grey button design + make it disabled:
Screenshot 2025-02-13 034319
- On click, the user gets a popup to use the extension popup instead:
Screenshot 2025-02-13 040247

Let me know what you think of these changes.

Step7750 (approved):
LGTM

webpack.config.js
```diff
@@ -21,13 +21,16 @@ function getPathEntries(path) {
...
         page: 'src/background_ff.html',
     };
     cp.browser_specific_settings = {
         gecko: {
             id: '{194d0dc6-7ada-41c6-88b8-95d7636fe43c}',
-            strict_min_version: '109.0',
+            strict_min_version: '127.0',
         },
     };
     // Allow getting the extension version from CSFloat page in Firefox
```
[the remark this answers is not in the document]
GODrums: That makes sense! Adjusted it accordingly in 4b702ee now.

From my experience the worst part is that the permission prompt appears right below the extension popup which makes it easy to miss. But that's an issue for another day :)

PR #300, opened 2025-03-17
Bluegem Data for Item Metadata and Inventory Pages

GODrums:
## Motivation

Case Hardened and Heat Treated are heavily pattern-based skin families. We want to provide users with additional data for each specific pattern.

## Description

This PR adds details for Case Hardened and Heat Treated items to Steam:
- Front / Top blue percentage as part of the metadata for inventory items & market pages
- More detailed text description with front / back or top / magazine (for AK) blue percentages

Also included is the script to generate the JSON file. Right now it requires Deno to be installed and can be executed via:

```
npm run generate_bluegem
```

## Screenshots

Item metadata in an inventory:
image

Large inventory items:
image

Item details on a market page:
image


scripts/generateBluegemJson.ts
```diff
@@ -0,0 +1,103 @@
+/// <reference lib="deno.ns" />
```
Step7750, line 1: Could we move this to `tools/generate_bluegem_json.ts`? `scripts` is somewhat of an overloaded term in a JS/TS repo.

src/lib/components/common/item_holder_metadata.ts
```diff
@@ -141,6 +154,24 @@ export abstract class ItemHolderMetadata extends FloatElement {
...
+                } catch (e: any) {
+                    console.error(`Failed to fetch bluegem for ${this.assetId}: ${e.toString()}`);
+                }
+            }
+        }
+    }
+
+    async annotateBluegem(info: ItemInfo) {
+        const bluegemResponse = await ClientSend(FetchBluegem, {
+            type: info.weapon_type!.replace(' ', '_'),
```
Step7750, line 170: nit: Generally I'd prefer indexing based on def_index instead of weapon type (ie. the string replacement is somewhat arbitrary).

However, if this makes it difficult to generate the JSON file with def_index as the key, then we can keep as is.

src/lib/components/common/item_holder_metadata.ts
```diff
@@ -141,6 +154,24 @@ export abstract class ItemHolderMetadata extends FloatElement {
...
+                try {
+                    await this.annotateBluegem(this.itemInfo);
+                } catch (e: any) {
+                    console.error(`Failed to fetch bluegem for ${this.assetId}: ${e.toString()}`);
+                }
+            }
+        }
+    }
+
+    async annotateBluegem(info: ItemInfo) {
```
Step7750, line 168: Annotate the type for the result.

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -13,12 +13,15 @@ import {
     floor,
```
Step7750, line 13: We should also annotate this for Steam Market listings.

Step7750 (changes requested):

src/lib/components/common/item_holder_metadata.ts
GODrums: I tend to agree, in this case I just took over the naming from the API we're getting the data from originally (aka. CSBlueGem). We would have to hard-code the def indices, but it's easy to change the script accordingly. I'll fix it tomorrow.

GODrums: The Five-Seven Heat Treated and Case Hardened have the same def_index, so this change seems to be a bit more complicated. We could move Heat Treated skins up in the range (like +1000), but we would need to make sure to remember this when using the data.
This feels like a big design decision, so what do you think?

src/lib/bridge/handlers/fetch_bluegem.ts
```diff
@@ -0,0 +1,44 @@
...
+    backside_purple?: number;
+    backside_gold?: number;
+    playside_contour_blue?: number;
+    playside_contour_purple?: number;
+    backside_contour_blue?: number;
+    backside_contour_purple?: number;
+}
+
+export interface FetchBluegemRequest {
+    type: string;
```
Step7750, line 18: Perhaps we should let them pass the `iteminfo` instead of the type + paintseed, so that we don't need to do string replacement in multiple places.

src/lib/components/common/item_holder_metadata.ts
Step7750: In that case, we'd need to define it with defindex and paintindex.

I'm fine with keeping it this way if we pass an iteminfo into the request and then perform the string type replacement once in the `FetchBluegem` handler.

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -147,6 +158,22 @@ export class SelectedItemInfo extends FloatElement {
...
+        if (!this.itemInfo || !this.bluegemData) {
+            return html``;
+        }
+
+        // Some skins got only one blue value
+        if (this.bluegemData.backside_blue === undefined) {
+            return html`<div>Blue: ${this.bluegemData.playside_blue}%</div>`;
+        }
+
+        const placement = this.itemInfo.weapon_type === 'AK-47' ? 'Top / Magazine' : 'Front / Back';
```
Step7750, line 171: We perform this check in a couple places -- perhaps the return value of the blue gem data should return the placement string.

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -203,6 +230,21 @@ export class SelectedItemInfo extends FloatElement {
             } catch (e: any) {
                 console.error(`Failed to fetch float for ${this.asset.assetid}: ${e.toString()}`);
             }
+
+            // Fetch bluegem data if needed
+            if (this.itemInfo && isBlueSkin(this.itemInfo)) {
```
Step7750, line 235: We should probably reset the value of `bluegemData` if this is false (or we fail to fetch below).

Otherwise, I think switching between two blue gems, where the second one fails to fetch below, would show incorrect previous blue gem data.

Step7750 (changes requested):

src/lib/components/common/item_holder_metadata.ts
GODrums: Fair! Changed it accordingly in f8250b8.

src/lib/components/common/item_holder_metadata.ts
```diff
@@ -141,6 +154,27 @@ export abstract class ItemHolderMetadata extends FloatElement {
 
         if (this.itemInfo) {
             this.annotateRankShine(this.itemInfo);
+
+            if (isBlueSkin(this.itemInfo)) {
+                try {
+                    await this.annotateBluegem(this.itemInfo);
```
Step7750, line 160: I think we can inline the `ClientSend` directly here.

```
this.bluegemData = await ClientSend(FetchBluegem, {
            iteminfo: info,
});
```

src/lib/components/common/item_holder_metadata.ts
```diff
@@ -141,6 +154,27 @@ export abstract class ItemHolderMetadata extends FloatElement {
 
         if (this.itemInfo) {
             this.annotateRankShine(this.itemInfo);
+
+            if (isBlueSkin(this.itemInfo)) {
```
Step7750, line 158: We should ensure we reset `this.bluegemData` if it isn't a blue skin anymore.

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -203,6 +230,23 @@ export class SelectedItemInfo extends FloatElement {
...
+                try {
+                    const bluegemResponse = await ClientSend(FetchBluegem, {
```
Step7750, line 237: nit: You can assign `this.bluegemData` directly from the response instead of this intermediate variable.

src/lib/components/market/item_row_wrapper.ts
```diff
@@ -130,6 +135,23 @@ export class ItemRowWrapper extends FloatElement {
             });
         }
 
+        // Fetch bluegem data if needed
+        if (this.itemInfo && this.asset && isBlueSkin(this.itemInfo)) {
+            try {
+                const bluegemResponse = await ClientSend(FetchBluegem, {
```
Step7750, line 141: nit: You can assign `this.bluegemData` directly from the response instead of this intermediate variable.

Step7750 (approved):

src/lib/components/common/item_holder_metadata.ts
GODrums: It shouldn't matter for the metadata components as the displayed skin never changes I think? I still added it for completeness now.

PR #306, opened 2025-03-19
Fix: Extension Overlap on Market Pages

GODrums:
## Description

When using the extension together with other popular extensions, this often causes overlaps and layout shifts.
This PR aims to **remove** overlapping elements, specifically the float, sticker and keychain containers on market pages. 

## Screenshots
Only CSFloat extension:
Screenshot 2025-03-19 at 19 04 58
With CSFloat + SIH:
Screenshot 2025-03-19 at 19 01 31
With CSFloat + CS2 Trader:
Screenshot 2025-03-19 at 19 01 14


src/lib/components/market/item_row_wrapper.ts
```diff
@@ -241,4 +244,32 @@ export class ItemRowWrapper extends FloatElement {
...
+
+    /**
+     * Avoid overlap with other extensions
+     */
+    cleanupOverlaps() {
+        // CS2 Trader
+        waitForTrue(() => $J(this).parent().parent().find('.stickerHolderMarket').length > 0).then((result) => {
+            if (!result) return;
+
+            const listingRow = $J(this).parent().parent();
```
Step7750, line 256: I think we could try to make a declarative decorator (similar to how we have the `@Inject` decorator).

Perhaps something like:
```
enum ConflictingExtension {
    CSGO_TRADER,  // Link to extension
    SIH, // Link to extension
}


@HideConflictingExtensionElem(ConflictingExtension.SIH, '#searchResultsRows .market_listing_row .market_listing_item_name_block .stickerHolderMarket', ConflictingMode.CONTINUOUS)
@HideConflictingExtensionElem(ConflictingExtension.SIH, '#searchResultsRows .market_listing_row .market_listing_item_name_block .stickersTotal', ConflictingMode.CONTINUOUS)
export class ItemRowWrapper extends FloatElement {
...
}
```

In the future, if we decide to make hiding conflicts optional, then it'd provide a central place to do it.

src/lib/components/market/item_row_wrapper.ts
```diff
@@ -241,4 +244,32 @@ export class ItemRowWrapper extends FloatElement {
...
+            listingRow.find('.stickerHolderMarket').css('display', 'none');
+            listingRow.find('.stickersTotal').css('display', 'none');
+            listingRow.find('.floatBarMarket').css('display', 'none');
+        });
+
+        // Steam Inventory Helper
+        waitForTrue(() => $J(this).parent().parent().find('div.sih-images').length > 0).then((result) => {
+            if (!result) return;
+
+            $J(this).parent().css({
```
Step7750, line 266: I'd imagine this is just un-doing the styles applied by SIH? If we wanted to do this in a more declarative way, that may be a bit annoying -- however, I guess a `@RemoveConflictingStyles('<selector>', ['prop1', 'prop2'])` may work.

Worst case, we could keep as-is, but I suspect this will come up in other places.

Step7750 (changes requested):

src/lib/components/market/item_row_wrapper.ts
GODrums: Yeah, SIH is applying a lot of additional CSS. In theory we can always apply this CSS as it's just Steam's default values (overwriting SIH overwrites). I'll try out the idea with decorators first though.

src/lib/components/market/item_row_wrapper.ts
GODrums: Great idea honestly! I created a more generic decorator that is able to cover both use cases in 4641b66 now.
Do let me know if this fits your initial idea.

src/lib/components/market/item_row_wrapper.ts
```diff
@@ -16,9 +16,24 @@ import {pickTextColour} from '../../utils/colours';
 import '../common/ui/floatbar';
 import {FetchBluegem, FetchBluegemResponse} from '../../bridge/handlers/fetch_bluegem';
 import {ClientSend} from '../../bridge/client';
+import {ConflictingMode, HideConflictingElement, StyleConflictingElement} from '../decorators';
 
 @CustomElement()
 @InjectAppend('#searchResultsRows .market_listing_row .market_listing_item_name_block', InjectionMode.CONTINUOUS)
+// Hide CS2 Trader elements
```
Step7750, line 23: nit: If we're planning to comment which extension it is anyways, I'd prefer to add it as an enum in the decorator. In the future, we may want to add further detection of whether they have CS2 Trader or SIH installed and only apply the hide conflicting element if they are installed (may be needed to prevent false positive hide/style).

Step7750 (approved):

src/lib/components/market/item_row_wrapper.ts
Step7750: Looks good, thanks!

src/lib/components/market/item_row_wrapper.ts
Step7750: LGTM

src/lib/components/market/item_row_wrapper.ts
GODrums: Fair! If we add that automatic detection in the future, it definitely makes sense to directly pass it as parameter here. Added it in ac0eaa5.

PR #311, opened 2025-03-28
Gradients for Acid / Amber Fades

GODrums:
## Description

This PR adds new gradients for the different types of already supported fade skins (i.e. Acid Fade & Amber Fade). Until now they used the typical fade gradient, which of course doesn't reflect their actual color scheme.
The new gradients are similar to the ones added to the website, but adapted for better visibility considering Steam's dark background and theme.

## Screenshots
Old gradients:
Screenshot 2025-03-28 at 13 22 11
New gradients:
Screenshot 2025-03-28 at 13 22 11

### DIFF
--- a/src/lib/utils/skin.ts
+++ b/src/lib/utils/skin.ts
@@ -165,7 +165,7 @@ export function isBlueSkin(itemInfo: ItemInfo): boolean {
 
 export function getFadeCalculatorAndSupportedWeapon(
     asset: rgAsset
-): [typeof FadeCalculator | typeof AcidFadeCalculator | typeof AmberFadeCalculator, string] | undefined {
+): [typeof FadeCalculator | typeof AcidFadeCalculator | typeof AmberFadeCalculator, string, string] | undefined {
     const FADE_TYPE_TO_CALCULATOR = {
         Fade: FadeCalculator,
         'Acid Fade': AcidFadeCalculator,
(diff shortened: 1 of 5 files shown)


src/lib/utils/skin.ts
```diff
@@ -175,19 +175,25 @@ export function getFadeCalculatorAndSupportedWeapon(
     for (const [fadeType, calculator] of Object.entries(FADE_TYPE_TO_CALCULATOR)) {
         for (const supportedWeapon of calculator.getSupportedWeapons()) {
             if (asset.market_hash_name.includes(`${supportedWeapon} | ${fadeType}`)) {
-                return [calculator, supportedWeapon.toString()];
+                return [calculator, supportedWeapon.toString(), fadeType.replace(' ', '-').toLowerCase()];
```
Step7750, line 178: I think we're starting to overload the term "Fade Type" -- we have both the paint name mapping above and this secondary form of "Fade Type" that is used for the class.

I'd prefer to make this a more explicit mapping and call it a different term (perhaps `class_name`). We're at a point where I think returning an object instead of an array is the way to go, especially with two back to back strings.

src/lib/utils/skin.ts
```diff
@@ -165,7 +165,7 @@ export function isBlueSkin(itemInfo: ItemInfo): boolean {
 
 export function getFadeCalculatorAndSupportedWeapon(
```
Step7750, line 166: `getFadeCalculatorAndSupportedWeapon` is outdated, we're also returning the `Fade Type` now.

Perhaps `getFadeParams` or similar to make this more generic.

Step7750 (changes requested):
:+1: Thanks @GODrums

src/lib/utils/skin.ts
```diff
@@ -175,19 +179,29 @@ export function getFadeCalculatorAndSupportedWeapon(
...
 
-    if (fadeCalculatorAndSupportedWeapon !== undefined) {
-        const [calculator, supportedWeapon] = fadeCalculatorAndSupportedWeapon;
+    if (fadeInfo !== undefined) {
+        const {calculator, weaponName, className} = fadeInfo;
 
-        return calculator.getFadePercentage(supportedWeapon, itemInfo.paintseed).percentage;
+        return {
+            percentage: calculator.getFadePercentage(weaponName, itemInfo.paintseed).percentage,
+            fadeType: className,
```
Step7750, line 203: We should avoid remapping this name to `fadeType`, I'd rather call this `className` here as well or define a `fadeType` enum with a function to get the class name on it.

Step7750 (approved):

PR #313, opened 2025-04-05
Chore: Update Deps and Build for Typescript 5

GODrums:
This PR updates our dependencies to match the newest versions (excluding Prettier & ESLint). This allows us to make use of various new performance improvements and the new Typescript 5 and Lit 3 features.

To accommodate to the new build process, paths have been updated with the `./`-prefix and Javascript `es2018`. New build has been tested to work on both Chrome and Firefox.

We also remove the `@types/cheerio` dependency, which isn't needed since since the removal of cheerio a year ago.


Step7750 (approved):
:+1:

PR #317, opened 2025-04-17
Fix: Release Note for Firefox

GODrums:
## Motivation

For successful extension releases via workflow on Firefox, we have to include a release note.

## Description

This PR adds a generic release note to the submit workflow for the Firefox extension.
The note links to the Github releases page to avoid copy-pasting the content of the release to the Firefox Store.

### DIFF
--- a/.github/workflows/submit_firefox.yml
+++ b/.github/workflows/submit_firefox.yml
@@ -42,5 +42,7 @@ jobs:
             approval-note: |
               The source code and build instructions are available at 
               https://github.com/csfloat/extension?tab=readme-ov-file#how-to-build-release.
+            release-note: |
+              Learn more about this release and its changes from the latest [Github release](https://github.com/csfloat/extension/releases).
             auth-api-issuer: ${{ secrets.AUTH_API_ISSUER }}
             auth-api-secret: ${{ secrets.AUTH_API_SECRET }}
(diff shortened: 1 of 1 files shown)


.github/workflows/submit_firefox.yml
```diff
@@ -42,5 +42,7 @@ jobs:
             approval-note: |
               The source code and build instructions are available at 
               https://github.com/csfloat/extension?tab=readme-ov-file#how-to-build-release.
+            release-note: |
+              Learn more about this release and its changes from the latest [Github release](https://github.com/csfloat/extension/releases).
```
Step7750, line 46: I think we should reference the version directly since otherwise it'll look a bit odd to see "latest" for every version on https://addons.mozilla.org/en-CA/firefox/addon/csgofloat/versions/

Looks like `robinraju/release-downloader` exposes `tag_name` and `release_name` (https://github.com/robinraju/release-downloader?tab=readme-ov-file#output-variables).

So we could do something like:

```
Learn more about release ${{ steps.release.outputs.release_name }} by visiting the [GitHub Release](https://github.com/csfloat/extension/releases/tag/${{ steps.release.outputs.tag_name }})
```

Step7750 (changes requested):

Step7750 (approved):

PR #319, opened 2025-04-21
Add Compression for `bluegem.json`

### LINKED ISSUE #318: `bluegem.json` Too Large for Firefox's Extension Validator
Currently uploads of the extension are failing to deploy on Firefox due to `data/bluegem.json` being >4MB. Here's some colour on it: https://discourse.mozilla.org/t/bug-with-uploading-large-files-in-addon/85407/4

We'll need to find a way to shrink this file down to <4MB.

Few thoughts on the top of my mind:

1) This file compresses very well since it has a bunch of repetitive strings -- we could store a compressed version and decompress it on the fly
2) If we want to avoid compression, we could rename the long-form fields like "playside_blue" to a abbreviated "ps_b" or similar.

@GODrums could you please take a look to see what's a good option here? Thanks!

GODrums:
## Motivation

Fix #318.

## Description

This PR adds build-time compression (via Webpack plugin) and run-time decompression (`pako`) for the `bluegem.json`. 
The original `bluegem.json` with 8.3MB is compressed down to `bluegem.json.gz` with 720KB.

The new `CompressionPlugin` required another change:
```
[tsl] ERROR
      TS2688: Cannot find type definition file for 'compression-webpack-plugin'.
  The file is in the program because:
    Entry point for implicit type library 'compression-webpack-plugin'
```
To fix it, I removed the `typeRoots` entry from the `tsconfig.json`.


Step7750 (approved):
Thanks @GODrums !

PR #288, opened 2025-02-02
Feature: Modal to allow users to list directly from inventory to CSFloat

IzaakPrats:
# [Issue #277 List from Inventory](https://github.com/csfloat/extension/issues/277)

This PR adds the ability to list items directly on CSFloat from the Steam inventory. Users can create both buy now and auction listings with a streamlined interface.

## Features
- Direct listing creation from Steam inventory
- Support for both Buy Now and Auction listing types
- Recommended price fetching from CSFloat*
- Price adjustment via input or percentage slider
- Configurable auction duration (1-14 days)
- Trade hold detection to prevent bad ux
- Success state with direct links to view listing/stall

## Implementation Details
- Added new `FetchRecommendedPrice` handler to get price suggestions from CSFloat
- Added `ListItem` handler to create listings via CSFloat API
- Created `ListItemModal` component with:
  - Buy Now and Auction modes
  - Item tradability validation

## Testing
No build / run errors

Screenshot 2025-02-03 at 11 53 22 AM

Buy Now Happy Path

https://github.com/user-attachments/assets/10c545d5-630d-4ec3-a4f6-7088acc5e2a2

Auction Happy Path

https://github.com/user-attachments/assets/de527c64-7c3c-4759-b5ab-772462bc1dec

Not Logged In

https://github.com/user-attachments/assets/4666843d-bcf1-40cd-be85-3ed59b3e19ff

Price Fetch Failed

https://github.com/user-attachments/assets/e12cab39-52fc-47a5-9909-8576253bee4c

PR #325, opened 2025-05-16
Dynamically Fetches User Fee for Listing Modal

Step7750:
Instead of hard-coding it to 2%, this PR dynamically fetches the fee.

Related to #288

### DIFF
--- a/src/lib/bridge/handlers/fetch_csfloat_me.ts
+++ b/src/lib/bridge/handlers/fetch_csfloat_me.ts
@@ -1,0 +1,41 @@
+import {SimpleHandler} from './main';
+import {RequestType} from './types';
+import {environment} from '../../../environment';
+
+export interface FetchCSFloatMeRequest {}
+
+export interface FetchCSFloatMeResponse {
+    actionable_trades: number;
+    has_unread_notifications: boolean;
+    pending_offers: number;
+    user: {
+        steam_id: string;
+        username: string;
+        avatar: string;
+        stall_public: boolean;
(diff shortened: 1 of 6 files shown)


src/lib/bridge/handlers/fetch_csfloat_me.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export interface FetchCSFloatMeError {
+    code?: string;
+    message?: string;
+}
+
+export const FetchCSFloatMe = new SimpleHandler<FetchCSFloatMeRequest, FetchCSFloatMeResponse>(
+    RequestType.FETCH_CSFLOAT_ME,
+    async (req) => {
+        return fetch(`${environment.csfloat_base_api_url}/v1/me`).then((resp) => {
```
GODrums, line 31: During testing this caused a lot of 403s. We should add `{ credentials: 'include' }` to this request.

Something similar happens for me with the stall request in `fetch_stall.ts`, which this modal reuses. CORS can be weird with requests executed via user action, so we might just want to add the credentials there as well.

GODrums (changes requested):
Looks good. Just a few issues with 403s we should tackle.

src/lib/bridge/handlers/fetch_csfloat_me.ts
Step7750: Ah interesting, I haven't been able to replicate this however there's no harm in adding it -- done.

GODrums (approved):

PR #327, opened 2025-05-17
QoL Features for the List Item Modal

GODrums:
## Motivation

The new listing modal only contains a small subset of features from the website. Users may want more customization options.

## Description

This PR adds new elements to the "list item"-modal to provide higher feature parity with the CSFloat website. This includes:
- a link to get to CSFloat's /sell-page in case someone doesn't want to use / doesn't trust the modal
- an informative banner linking CSFloat's blog article for auctions (similar to `/sell`)
- an assessment label for the price (similar to `/sell`)
- an option to create a private listing
- a "search on market"-link to refer back to the market listings website
- option to add a description to the listing (hidden behind a plus per default)

## Screenshot
image
After clicking the "plus" for the description:
image


src/lib/components/inventory/list_item_modal.ts
```diff
@@ -8,11 +8,14 @@ import {ClientSend} from '../../bridge/client';
 import {ListItem} from '../../bridge/handlers/list_item';
 import {FetchRecommendedPrice} from '../../bridge/handlers/fetch_recommended_price';
 import {listItemModalStyles} from './list_item_modal_styles';
-import {isLoggedIntoCSFloat} from '../../utils/auth';
 import {CSFError, CSFErrorCode} from '../../utils/errors';
-import {FetchPendingTrades} from '../../bridge/handlers/fetch_pending_trades';
 import {FetchCSFloatMe} from '../../bridge/handlers/fetch_csfloat_me';
 
+// SVG icons for percentage assessment
```
Step7750, line 14: Perhaps we can move these icons to a central `icons.ts` file either in this directory or in a more global place?

src/lib/components/inventory/list_item_modal.ts
```diff
@@ -354,17 +382,20 @@ export class ListItemModal extends FloatElement {
...
                                 </a>
                                 <a
                                     href="https://csfloat.com/stall/me"
                                     target="_blank"
-                                    class="base-button secondary-button success-link"
+                                    class="base-button secondary-button"
                                 >
                                     View Your Stall on CSFloat
                                 </a>
+                                <a href="${this.searchUrl}" target="_blank" class="base-button secondary-button">
```
Step7750, line 396: They may want a way to search these items without having to list the item first -- perhaps there's a way to put a UI element somewhere (maybe to the right of the recommended price above the input).

src/lib/components/inventory/list_item_modal.ts
```diff
@@ -468,15 +544,44 @@ export class ListItemModal extends FloatElement {
...
+                                      </div>
+                                      <span class="percentage-assessment-value"> ${pricePercentage}% </span>
+                                  </div>
+
                                   <div>
-                                      Percentage of recommended price:
-                                      ${this.recommendedPrice && this.customPrice
-                                          ? Math.round((this.customPrice / this.recommendedPrice) * 100)
-                                          : 100}%
+                                      <label>Visibility</label>
```
Step7750, line 566: Toggling visibility to private is pretty rare -- this UI component takes a large amount of space given how infrequent it'll be used.

Perhaps we can right-align it on the same line as the "Visibility" text and make the buttons much smaller?

Step7750 (changes requested):
Thanks for these QOL features @GODrums !

GODrums:
Additionally, I added listing support for the new music kits and pins now in 74e0e03.
And it seemed like the charms-detection was broken as well. With 10ac9ad it should be fixed again.

GODrums:
Added a new description field now. Per default this it is hidden behind a "plus"-button to avoid overwhelming the user.

I also fixed another issue from the original modal PR, which caused the slider to not be updated when manually entering a price (35a60b4).

Step7750 (approved):
This is great @GODrums ! LGTM

PR #330, opened 2025-05-21
Update "List on CSFloat" Logo

GODrums:
Screenshot 2025-05-22 at 00 51 21

### DIFF
--- a/src/lib/components/inventory/selected_item_info.ts
+++ b/src/lib/components/inventory/selected_item_info.ts
@@ -45,8 +45,8 @@ export class SelectedItemInfo extends FloatElement {
                 margin: 10px 0 10px 0;
                 padding: 5px;
                 width: fit-content;
-                border: 1px #5a5a5a solid;
-                background-color: #383838;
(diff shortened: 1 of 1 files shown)


src/lib/components/inventory/selected_item_info.ts
```diff
@@ -187,7 +187,7 @@ export class SelectedItemInfo extends FloatElement {
         return html`
             <div class="market-btn-container">
```
Step7750, line 188: Perhaps we can update the `market-btn-container` to match the color palette.

Played around a little bit with `border: solid 1px rgb(56 64 77)` and  `background-color: rgb(43 48 57)` and got:

image


The background does match the logo quite closely, but I don't have a strong preference.

Step7750 (changes requested):

src/lib/components/inventory/selected_item_info.ts
GODrums: Agreed, these fit the modal a bit better as well. Changed in 392fc92.

PR #336, opened 2025-07-18
Fix: Handle New Highlight Charms

GODrums:
Currently, the extension displays an empty pattern ID for highlight / souvenir charms as they are treated like regular charms.
image
image

 This PR handles them explicitly as non-sellable on CSFloat and removes the pattern.

### DIFF
--- a/src/lib/components/market/item_row_wrapper.ts
+++ b/src/lib/components/market/item_row_wrapper.ts
@@ -138,6 +138,11 @@ export class ItemRowWrapper extends FloatElement {
             inlineEasyInspect($J(this).parent().parent().find('.market_listing_item_img_container'), this.inspectLink);
         }
 
+        // If the item is a highlight charm, we're done here
+        if (this.asset && isHighlightCharm(this.asset)) {
+            return;
+        }
+
         try {
             this.itemInfo = await this.fetchFloat();
         } catch (e: any) {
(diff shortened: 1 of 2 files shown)


src/lib/utils/skin.ts
```diff
@@ -128,6 +128,10 @@ export function isCharm(asset: rgAsset): boolean {
     return isAbstractType(asset, 'Charm', 'CSGO_Tool_Keychain');
 }
 
+export function isHighlightCharm(asset: rgAsset): boolean {
+    return isCharm(asset) && asset.tags!.some((a) => a.internal_name === 'highlight');
```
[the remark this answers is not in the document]
GODrums: the isCharm function guarantees that the tags exists, so this is fine.

src/lib/components/market/item_row_wrapper.ts
```diff
@@ -239,7 +248,7 @@ export class ItemRowWrapper extends FloatElement {
                     ${this.renderBluegem()}
                 </div>
             `;
-        } else if (this.itemInfo && isCharm(this.asset)) {
+        } else if (this.itemInfo && isCharm(this.asset) && !isHighlightCharm(this.asset)) {
```
Step7750, line 251: We should probably also block fetching the item info at all (it's a needless call that can't be permanently cached like skins). We could do this in a follow up however, if you'd like.

src/lib/utils/skin.ts
Step7750, line 132: Let's be a bit safe here and check that `tags` is defined.

Step7750 (approved):

src/lib/components/market/item_row_wrapper.ts
GODrums: Good point! With 6c88d13 there should be no more `itemInfo` fetches for Highlight Charms anymore, including in `selected_item_info`.
Do let me know if this change satisfies your suggestion.

PR #341, opened 2025-08-17
[src/lib/filter] Add New Method seedInList

albertjsung:
**Summary**
--- 
`seedInList` allows users to filter for items that satisfy a list of paint seeds they are looking for. This is particularly useful for people who are looking for tier1/tier2/tier3 patterns, which each have 30+ paint seeds and can be quite tedious to search for.

E.g. seed == 301 or seed == 625 or ... or seed == 540 gets annoying pretty quick and most people do manual cmd/crtl + f searches or use a custom script.

Examples:
t1_bayonet_gamma_doppler_p1
snow_leopard_t1

**Implementation Details + Demo (ran dev build in chrome)**
---
Most pattern guides for CS2 are made by korenevskiy, and those that aren't mostly follow his formatting. 

I've classified the two most common ways guides list the pattern seeds:

1. seed1 seed2 ... seed10
2. seed1, seed2, ..., seed10

The implementation will allow us to **directly copy and paste the numbers from the images above** and pass it into our new method.

Example One (most common case, seeds separated by seeds):
demo

Example Two (with commas this time):
demo2

**Future Work**
---
Once approved, will need to add this method to `filter_help.ts` so users can see and understand how to use this method properly. Can probably find a better name for this method as well so it's more intuitive to users.

Also, this feature would be extremely useful for the csfloat market search filters as well. But I'm not sure how they go about receiving feature ideas from the public.

### DIFF
--- a/src/lib/filter/custom_functions.ts
+++ b/src/lib/filter/custom_functions.ts
@@ -23,6 +23,9 @@ export function match(str: string, regex: string) {
     else return 0;
 }
 
-export function seedInList(seed: number, seedList: number[]) {
-    return seedList.includes(seed);
+export function seedInList(seed: string | number, seedList: (string | number)[]) {
+    const numericSeed = Number(seed);
+    const numericList = seedList.map(Number);
+
+    return numericList.includes(numericSeed);
 }
--- a/src/lib/filter/custom_functions.ts
+++ b/src/lib/filter/custom_functions.ts
@@ -25,8 +25,14 @@ export function match(str: string, regex: string) {
 
 export function seedInList(seed: string, seeds: string) {
     if (typeof seeds !== "string") {
-        throw new TypeError(`seedGroup must be a string, got ${typeof seeds}`);
+        throw new TypeError(`seeds must be a string, got ${typeof seeds}`);
     }
+    
+    if (typeof seed !== "string" && typeof seed !== "number") {
+        throw new TypeError(`seed must be a string or number, got ${typeof seed}`);
+    }
+
+    const seedStr = seed.toString();
 
     const seedList = seeds
         .replace(/,/g, ' ')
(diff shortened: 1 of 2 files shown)


src/lib/filter/custom_functions.ts
```diff
@@ -22,3 +22,18 @@ export function match(str: string, regex: string) {
...
+
+export function seedInList(seed: string | number, seedList: (string | number)[]) {
+    if (!Array.isArray(seedList)) {
+        throw new TypeError(`seedList must be an array, got ${typeof seedList}`);
+    }
+
+    const numericSeed = Number(seed);
+    if (Number.isNaN(numericSeed)) return false;
+
+    const numericList = seedList
```
albertjsung, line 34: Could also consider using a singular set of seeds and checking if each market listing's paint seed has membership in the set. This would make lookups faster, but also logic is a bit more complicated and didn't want to get ahead of myself.

For what it's worth this feature could already exist and I didn't look hard enough.

src/lib/filter/custom_functions.ts
```diff
@@ -22,3 +22,15 @@ export function match(str: string, regex: string) {
     if (thisMatch !== null) return thisMatch.length;
     else return 0;
 }
+
+export function seedInList(seed: string, seeds: string) {
+    if (typeof seeds !== "string") {
+        throw new TypeError(`seedGroup must be a string, got ${typeof seeds}`);
+    }
+
+    const seedList = seeds
```
albertjsung, line 31: Could also consider using a singular set of seeds and checking if each market listing's paint seed has membership in the set. This would make lookups faster, but implementation is a bit more complicated and didn't want to get ahead of myself.

For what its worth this feature could already exist and I didn't look hard enough.

PR #345, opened 2025-09-09
Prevents Sending Redundant Trade Pings as Seller

Step7750:
Re-affirming the state of a trade offer or trade history as a seller doesn't provide value server-side.

Will extend to buyers as well in the future, but it requires a couple server-side changes in the works.


src/lib/alarms/trade_history.ts
```diff
@@ -1,27 +1,45 @@
...
         const received_ids = e.received_assets.map((e) => e.asset_id);
         const given_ids = e.given_assets.map((e) => e.asset_id);
-        return !![...received_ids, ...given_ids].find((e) => {
-            return assetsToFind[e];
-        });
+
+        const foundSlimTrades = [...received_ids, ...given_ids]
+            .map((e) => {
+                return assetsToFind[e];
+            })
```
GODrums, line 29: nit: `.map((e) => assetsToFind[e])` is a bit more compact

GODrums (approved):

PR #387, opened 2026-04-13
Implements FloatDB Gateway for Rank Fetching

Step7750:
## Overview

Switches to using the `gateway.floatdb.com` service to lookup ranks for items. There's been a decent chunk of optimization to reduce needless calls, such as using a threshold blob to determine whether an item qualifies for a rank.

## Changes
* Pre-filters rank lookups with a locally cached threshold copy
* Eagerly fetches ranks when possible to reduce the amount of total calls to the gateway
  * Note: This only applies when visiting a Steam profile inventory, notably excluding the trade inventory page
  * Batches rank lookups within 100ms debouncing periods to similarly prevent many calls
* Refactors implementation of the float fetcher service since it doesn't need a queue model anymore
  * Fixes previous issues with deduplicating in flight requests and returning cached items
* Switches to primarily use a new `FetchInspectInfoBatch` bridge method since the overhead of sending 100+ messages to the service worker to check ranks would exceed the rank batcher debounce interval.

Ref CSF-1165

---

> [!NOTE]
> **Medium Risk**
> Introduces new networked rank-check and threshold-fetch flows plus request batching/deduping, which could affect item metadata/rank display correctness and increase reliance on the new gateway endpoints.
> 
> **Overview**
> Switches rank lookups to the FloatDB Gateway (`environment.floatdb_gateway_url`) and adds client-side rank-check batching plus local threshold caching to reduce unnecessary rank requests.
> 
> Refactors float/inspect fetching to require `asset_id` and to batch bridge calls via a new `FETCH_INSPECT_INFO_BATCH` handler; `FloatFetcher` now microtask-batches, de-dupes in-flight requests, and caches results by asset.
> 
> Adds eager prefetching of floats/ranks when inventory asset properties load, and updates market/inventory components to pass `asset_id` (removing the previous list-price-based fetch path).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8204d10022298ff2a298ce862d54e867cddb0b63. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/environment.ts
```diff
@@ -6,4 +6,5 @@ export const environment = {
         loggingLevel: 'Warn',
     },
     reverse_watch_base_api_url: 'https://reverse.watch/api',
+    floatdb_gateway_url: 'https://gateway.floatdb.com',
```
GODrums: Since this is a new domain, we'd need to request host permissions for it, otherwise requests will fail in production.

In turn this will mean the extension will get disabled for all users until they accept that permission

src/lib/bridge/handlers/fetch_inspect_info.ts
```diff
@@ -42,95 +44,144 @@ export interface ItemInfo {
...
+        weapon_type: weaponType,
+        item_name: itemName,
+        rarity_name: rarityName,
+        wear_name: getWearName(floatvalue),
+    };
+
+    try {
+        if (decoded.itemid != null && decoded.paintwear != null) {
+            const stattrak = decoded.killeaterscoretype !== undefined;
+            const souvenir = (decoded.quality ?? 0) === 12;
```
GODrums:
```suggestion
            const souvenir = decoded.quality === 12;
```

src/lib/storage/keys.ts
```diff
@@ -59,3 +60,7 @@ export const SCHEMA_CACHE = newRow<{
     lastUpdated: number;
     schema: ItemSchema.Response;
 }>(StorageKey.SCHEMA_CACHE);
+export const THRESHOLD_CACHE = newRow<{
+    lastUpdated: number;
+    thresholds: Record<string, {low: number; high: number}>;
```
GODrums: we should reuse our `ThresholdEntry` type here. 
Of note, this may require `import type` to avoid a circular dependency

GODrums (approved):
🥇

src/lib/page_scripts/inventory.ts
```diff
@@ -1,7 +1,69 @@
...
+     */
+    Observe(
+        () => {
+            const count = Object.keys(getAllCS2AssetProperties()).length;
+            if (count > 0 && !initialRun) {
+                initialRun = true;
+                return true;
+            }
+            return count;
+        },
```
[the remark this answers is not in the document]
Step7750: This is intentional, there is no downside to requesting the same item twice since it is cached/deduped.

src/lib/services/rank_batcher.ts
```diff
@@ -0,0 +1,82 @@
...
+
+            if (!resp.ok) {
+                throw new Error(`Ranks check failed with status ${resp.status}`);
+            }
+
+            const data: {ranks: Record<string, RankResult>} = await resp.json();
+
+            for (const [assetId, item] of batch) {
+                const rank = data.ranks[assetId] ?? null;
+                item.deferred.resolve(rank);
```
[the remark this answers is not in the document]
Step7750: Sorry Cursor, but this one is a miss -- the link encodes the asset ID.

Issue #395, opened 2026-06-04
Create Architecture to Handle Steam Market Beta


GODrums:
Adapting the extension for the new Steam Market Beta is a multi-step process. We want to ensure we continue to provide backwards compatibility for users, who opted out of the beta.

This task tackles the following areas:
- **Removal of JQuery**: Since the new beta doesn't inject JQuery anymore, some of our shared functionality has to be rewritten with common DOM APIs
- **State management**: The elimination of global variables in the `window` context means we need an alternative to keep in sync with the website’s internal state and data
- **Randomized class names**: The lack of consistency in class names means we have to find different ways to uniquely identify parts of the UI
- **Self CSP**: React applies the CSP `default-src: 'self'`, hence loading our own font sheet from Google or executing created inline functions won’t be possible anymore

PR #399, opened 2026-06-05
Integrate Fiber State Management and Float Ranks

### LINKED ISSUE #398: Restore listing enhancements in the market beta using a React Fiber compatible state management integration

GODrums:
Closes https://github.com/csfloat/extension/issues/398

Step 2 of https://github.com/csfloat/extension/issues/395

Adds state management integration with Fiber and first Fiber-based listing enhancement: float ranks. More enhancements to follow separately.

## Screenshot
image

---

> [!NOTE]
> **Medium Risk**
> Relies on undocumented React Fiber/DOM structure and brittle card selectors that may break when Steam updates the beta UI; changes are gated to the beta market only.
> 
> **Overview**
> Adds **React Steam Market (beta)** support by reading listing data from React Fiber instead of legacy `g_rgListingInfo` / `g_rgAssets`.
> 
> A new **`getFiberProps`** helper walks the `__reactFiber$` tree to pull typed **`MarketListing`** props (listing id, asset, inspect link, wear from asset properties). **`BetaListingEnhancer`** continuously injects into beta listing cards when `isReactSteamMarket()` is true, fetches inspect **`ItemInfo`** via the existing float fetcher, and mounts **`BetaListingRank`** beside the wear value span (matched by float text). The market listing page script now loads this enhancer alongside the legacy **`ItemRowWrapper`** path.
> 
> Only a trivial newline fix in **`mode.ts`**; legacy market behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 48bc03368edbdd4b98410ef1d7c3cf27b3f05815. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/lib/components/market/react/rank.ts
```diff
@@ -0,0 +1,60 @@
...
+    private findWearSpan(): HTMLSpanElement | null {
+        if (this.targetFloat === null) return null;
+
+        const spans = this.card.querySelectorAll<HTMLSpanElement>('span[style*="pre-wrap"]');
+        for (const span of spans) {
+            const text = span.textContent?.trim();
+            if (!text) continue;
+            const value = parseFloat(text);
+            if (!Number.isNaN(value) && Math.abs(value - this.targetFloat) < 1e-6) return span;
+        }
```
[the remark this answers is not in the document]
GODrums: From my testing at least, we can reasonably assume that these aren't rounded or formatted at all, Steam uses the `eclipse` overflow styling here

src/lib/components/market/react/rank.ts
```diff
@@ -0,0 +1,60 @@
...
+    private placeNextToWear(): void {
+        if (this.injected || !this.itemInfo || !this.card) return;
+        if (!parseRank(this.itemInfo)) return;
+
+        const wearSpan = this.findWearSpan();
+        if (!wearSpan) return;
+
+        this.injected = true;
+        wearSpan.insertAdjacentElement('afterend', this);
+    }
```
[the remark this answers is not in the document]
GODrums: this renders late enough that this doesn't matter. If this is an item without wear, this component will never get rendered anyways

src/lib/components/market/react/listing.ts
```diff
@@ -0,0 +1,121 @@
...
+        } catch (e) {
+            return;
+        }
+
+        this.injectRank(info);
+    }
+
+    private injectRank(info: ItemInfo): void {
+        if (this.rankInjected) return;
+        this.rankInjected = true;
```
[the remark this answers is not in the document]
GODrums: This isn't wrong, but totally not needed. React's Fiber reconciler only ever operates on host nodes it created. The case of the card (parent element of the rank) unmounting is covered by the `BetaListingEnhancer` already, in which case this injection runs again.

Step7750: Is it necessary to track `rankInjected`? AFAIK, `connectedCallback()` shouldn't fire more than once for a given attachment. Is there a specific case where this is triggered?

Empirically, removing this `rankInjected` handling seems to produce no mis-fire or repeated injection.

src/lib/components/market/react/rank.ts
```diff
@@ -0,0 +1,60 @@
...
+
+        const spans = this.card.querySelectorAll<HTMLSpanElement>('span[style*="pre-wrap"]');
+        for (const span of spans) {
+            const text = span.textContent?.trim();
+            if (!text) continue;
+            const value = parseFloat(text);
+            if (!Number.isNaN(value) && Math.abs(value - this.targetFloat) < 1e-6) return span;
+        }
+        return null;
+    }
```
Step7750: Totally understandable that it is getting quite difficult to figure out which element relates to wear considering we can't rely on DOM siblings being consistent (i.e. some skins have elements like stickers).

However, we're starting to break the pattern of declarative handling (granted, I think some usages of the old market UI also did).

I'm wondering if we can find a middle-ground where `InjectionGuard` inside of `@Inject*` can provide some of this context filtering that can't easily be encoded into a query selector. For instance, we'd have a `@InjectAppend` decorator for `BetaListingRank` that attempts to target the exact element we want to append to. Since we expect our query selector to collide, we can pass the proposed element into the `InjectionGuard` and do some context handling to deduce whether a float is there.

As an alternative, to make our lives simpler, we could try to inject into a more stable element. However, you'd lose the flair of having the rank next to the float.

src/lib/components/market/react/rank.ts
```diff
@@ -0,0 +1,60 @@
+import {css, html, nothing} from 'lit';
+import {property} from 'lit/decorators.js';
+
+import {CustomElement} from '../../injectors';
+import {FloatElement} from '../../custom';
+import {ItemInfo} from '../../../bridge/handlers/fetch_inspect_info';
+import {parseRank, renderClickableRank} from '../../../utils/skin';
+
+@CustomElement()
+export class BetaListingRank extends FloatElement {
```
Step7750: nit: `Beta*` for a naming scheme here will get outdated -- perhaps we can have a prefix more related to the new setup? i.e. `ReactListingRank`

I guess eventually the old Steam Market will get purged and we'll likely strip this prefix anyways, so it's your call in the meantime.

src/lib/components/market/react/listing.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    private get card(): HTMLElement {
+        const parent = this.parentElement;
+        if (!parent) throw new Error('Card element not found');
+        return parent;
+    }
+
+    /** Steam implementation detail: the "key" property on the Fiber node is the listing ID. */
+    get fiberProps(): MarketListingProps | null {
+        return getFiberProps<MarketListingProps>(this.card, (fiber) => typeof fiber.key === 'string') ?? null;
```
Step7750: nit: I'd expect this to be more stable if we checked that `memoizedProps` contains `listing` as a field. Valve could add a ancestor element that is closer with some props that breaks this parsing more easily IMO.

Step7750 (changes requested):
Apologies for the late review!

The fiber parsing is quite nifty, thanks for figuring all that out -- certainly seems required with the lack of context when injecting.

I'm thinking we can potentially move this PR to be more declarative instead of injecting a wrapper that then finds a child element to manually inject into. This would require expanding the `InjectionGuard` to a degree, but IMO would enable us to directly configure injecting the rank element.

src/lib/components/market/react/listing.ts
GODrums: Fair, this is a bit of leftover from a previous version. Removed it now.

src/lib/components/market/react/listing.ts
GODrums: Not quite. The way this traversal works is that we traverse through the original React layers until we reach the base component. From my analysis, their code looks something like this:
```ts
function ListingCard({ listing, ... }) {
  return (
    <{/* Some memo/forwardRef wrapper */}>
      <Grid ...>
        <Focusable {/* React.context wrapper for controller / Steam Deck */} ...>
          <div ...> /* ← this is our entry HTML traversal element */
             {/* inner content */}
          </div>
        </Focusable>
      </Grid>
    </{wrapper}>
  );
}
```

So we basically traverse the non-rendered layers up until we reach the base of the React component. And we know this one will always contain `key` as it's the unique identifier of the list rendering. Something like this:
```ts
function ListingResults({ listings }) {
  return (
    <>
      {listings.map((listing) => (
        <ListingCard
          key={listing.listingid}        // ← this is fiber.key
          listing={listing}
          ...
        />
      ))}
    </>
  );
}
```

Technically, we could switch to detecting `listing` within `memoizedProps`, it shouldn't really matter at all though as there cannot be any React components in our traversal up before `key` is defined. On the other hand, if Valve decides to rename the `listing` props later on, we would traverse up until the top root in the worst case, which would make the website pretty laggy for users until we push a fix. Using `key` we are guaranteed to stop at the list-rendering React component.

src/lib/components/market/react/rank.ts
GODrums: Agreed, the imperative handling isn't ideal. I do believe using `ReactListingEnhancer` as orchestrator is the simplest and most stable approach though.

Without the context from `ReactListingEnhancer` (i.e. knowing the float we look for), the guard could only search for "a parseable number between 0 and 1", which becomes extremely brittle (e.g. a skin may have an applied name tag with a matching number).
This approach will quickly hit its limits, for example for our paint seed enhancements in https://github.com/csfloat/extension/pull/401 ("any number between 0 and 1000") as it wouldn't be able to distinguish between the pattern template and paint seed.

We also have to pass down the fetched listing props and `itemInfo` from `ReactListingEnhancer` (here: `low_rank`, `high_rank`) as all smaller injections depend on it. Fetching them directly wouldn't be possible.

I would leave the final choice up to you, but the declarative approach would definitely come with much stricter limits along the way, which might not be easily solvable.

src/lib/components/market/react/listing.ts
Step7750: Fair enough, thanks for the detail! We can keep as-is.

src/lib/components/market/react/rank.ts
Step7750: Perhaps we can define re-usable "scopes" that provide the anchor points that other elements can build off of.

For instance, we could define an injection scope like so:

```js
export const ReactMarketListingScope = defineInjectionScope<ReactListingContext>({
    selector: 'div[style*="--grid-rows"]:has([style*="market_listings/"])',
    mode: InjectionMode.CONTINUOUS,
    guard: isReactSteamMarket,
    context: buildReactListingContext,
});
```

and use it repeatedly for elements that want to inject like:

```js
@CustomElement()
@InjectIntoScope(ReactMarketListingScope, {
    anchor: findWearSpan,
    position: InsertPosition.After,
})
export class ReactListingRank extends FloatElement {}
```

I got AI to put together how this might work: https://github.com/csfloat/extension/commit/b4fd6331adb66f0ab21f60262b9c44134ae14dc7

I expect most of Steam to move over to React, so I'd really like to make this system simple and extensible to use.

Step7750 (changes requested):

src/lib/components/market/react/rank.ts
GODrums: Wow, that's a crazy approach, but I like it. Agreed that those under-the-hood scopes are much cleaner.

The main concern I see here is regarding performance as every new enhancement requires its own new selector interval. Probably not a big deal for now, but might be worth to keep in mind if we want to add many more enhancements.

Cherry picked your commit onto this branch now (62ed48e).

Step7750 (approved):
👍

PR #403, opened 2026-06-13
Restore Market Filter with Custom Expression Parser

### LINKED ISSUE #400: Restore filtering / highlighting logic by reworking our filter with our own AST parser (circumvent React's CSP default-src 'self')

GODrums:
Closes csfloat/extension#400

Step 3 of csfloat/extension#395

Adds a compatibility layer to support our filter container in the Steam Market Beta.

Due to React's `self` CSP, it also replaces `filtrex` with a custom expression parser supporting the same input syntax. Architected with Opus 4.8, reviewed by Fable 5. Tested all variables and operators manually

## Screenshots

!https://github.com/user-attachments/assets/1d296ca3-3ec6-4eed-88c5-7ec2f2d5bdef

---

> \[!NOTE\]<br>**Medium Risk**<br>Replaces the entire filter expression engine and broadens valid return types, so subtle syntax or truthiness mismatches vs filtrex could affect highlights across classic and beta market pages.
>
> **Overview**<br>Brings **CSFloat highlight filters** back on the **Steam Market beta** by mounting the existing filter UI above the React listing grid and wiring beta listing cards to the same `gFilterService` highlighting used on the classic market.
>
> `filtrex` **is removed** and replaced with an in-tree `compile_expression` tokenizer/parser/evaluator so filter syntax still works under React’s stricter CSP. `Filter` now compiles through that module; `isValidReturnValue` treats any non-`Error` result as valid instead of only booleans and `0`/`1`.
>
> Beta `BetaListingEnhancer` derives price from listing fiber props, subscribes to filter updates, and paints each card’s background/text when a filter matches.
>
> Reviewed by [Cursor Bugbot](<https://cursor.com/bugbot>) for commit 47b0317e43d38505ccb268b59c65bb40ba81730e. Configure [here](<https://www.cursor.com/dashboard/bugbot>).

### DIFF
--- a/src/lib/components/market/react/filter_panel.ts
+++ b/src/lib/components/market/react/filter_panel.ts
@@ -1,0 +1,61 @@
+import {css, html, nothing, HTMLTemplateResult} from 'lit';
+import {CustomElement, InjectBefore, InjectionMode} from '../../injectors';
+import {FloatElement} from '../../custom';
+import {isReactSteamMarket} from '../mode';
+
+import '../../filter/filter_container';
+
+/**
+ * Steam Market Beta equivalent of {@link UtilityBelt}.
+ */
+@CustomElement()
+@InjectBefore(
+    'div:has(> [style*="grid-columns:repeat(auto-fill, minmax(260px"])',
+    InjectionMode.CONTINUOUS,
+    isReactSteamMarket
+)
+export class BetaFilterPanel extends FloatElement {
+    static styles = [
+        ...FloatElement.styles,
+        css`
+            .panel {
+                margin-bottom: 16px;
+                padding: 16px;
+                background-color: rgba(0, 0, 0, 0.25);
+                border: 1px solid rgba(255, 255, 255, 0.06);
+                border-radius: 8px;
+                color: #c7d5e0;
+                font-family: 'Motiva Sans', sans-serif;
+            }
+
+            .panel-title {
+                font-family: 'Motiva Sans', sans-serif;
+                font-size: 14px;
+                font-weight: 600;
+                letter-spacing: 0.04em;
+                text-transform: uppercase;
+                color: #ebebeb;
+                margin: 0 0 12px;
+            }
+        `,
+    ];
+
+    private get key(): string {
+        return marketHashName() ?? '';
+    }
+
+    protected render(): HTMLTemplateResult | typeof nothing {
+        if (!this.key) return nothing;
+        return html`
+            <div class="panel">
+                <h3 class="panel-title">CSFloat Filters</h3>
+                <csfloat-filter-container .key="${this.key}"></csfloat-filter-container>
+            </div>
+        `;
+    }
+}
+
+/** Use the title of the page to get the market hash name */
+function marketHashName(): string | undefined {
+    return document.title.match(/^(.+?) - Steam Community Market$/)?.[1] ?? undefined;
+}
--- a/src/lib/components/market/react/filter_panel.ts
+++ b/src/lib/components/market/react/filter_panel.ts
@@ -41,7 +41,7 @@ export class ReactFilterPanel extends FloatElement {
     ];
 
     private get key(): string {
-        return marketHashName() ?? '';
+        return getSteamMarketID() || '';
     }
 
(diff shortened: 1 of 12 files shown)


src/lib/components/market/react/listing.ts
```diff
@@ -84,6 +88,15 @@ export class BetaListingEnhancer extends FloatElement {
...
     }
 
+    get convertedPrice(): number | undefined {
+        const listing = this.listing;
+        if (!listing || !listing.unPrice) {
+            return;
+        }
+
+        return (listing.unPrice + listing.unFee) / 100;
+    }
```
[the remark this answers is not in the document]
GODrums: since we use React's internal values here, no currency conversion is needed. Tested with EUR & USD in both logged in and logged out

src/lib/components/market/react/filter_panel.ts
```diff
@@ -0,0 +1,61 @@
...
+                <csfloat-filter-container .key="${this.key}"></csfloat-filter-container>
+            </div>
+        `;
+    }
+}
+
+/** Use the title of the page to get the market hash name */
+function marketHashName(): string | undefined {
+    return document.title.match(/^(.+?) - Steam Community Market$/)?.[1] ?? undefined;
+}
```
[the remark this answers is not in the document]
GODrums: replaced with Steam's new market ID format via the URL in 86a7466

src/lib/filter/compile_expression.ts
```diff
@@ -0,0 +1,665 @@
+import type {InternalInputVars} from './types';
```
Step7750, line 1: Unsure if we have the right infra for testing in the extension, but this would certainly be a spot to use it!

Not a hard blocker, but if it's a few prompts away then I'd say it's worth it.

src/lib/components/market/react/highlight.ts
```diff
@@ -0,0 +1,49 @@
...
+        if (!listing?.unPrice) return undefined;
+        return (listing.unPrice + listing.unFee) / 100;
+    }
+
+    private applyColour(): void {
+        const card = this.parentElement;
+        const context = this.injectionContext;
+        if (!card || !context) return;
+
+        const colour = gFilterService.matchColour(context.itemInfo, this.convertedPrice) || '';
```
Step7750: nit: If we don't have a color to apply, we should skip applying a style below. Otherwise, there is risk that we reset a bg color that Valve applies in the future.

src/lib/filter/compile_expression.ts
```diff
@@ -0,0 +1,665 @@
...
+        if (this.match('paren', '(')) {
+            const node = this.parseOr();
+            this.expect('paren', ')');
+            return node;
+        }
+
+        throw new Error(`unexpected token "${token.value}" at ${token.position}`);
+    }
+
+    /** Parses the `(...)` operand of `in` / `not in` into a list node. Allows an empty list `()`. */
```
Step7750: nit: empty list shouldn't be valid syntax IMO. Is there a use case where `var in ()` could be true?

Step7750 (approved):
Great stuff, the expression parser is quite clean and `ReactListingHighlight` is cleaner than I thought it'd be under the React framework.

src/lib/filter/compile_expression.ts
GODrums: Fantastic idea, added Vitest with a minimal testing setup (d05d7bf). `compile_expression.test.ts` contains a basic test suite for the parser, nothing too fancy (bafdfa1).

src/lib/components/market/react/highlight.ts
GODrums: This actually fulfills the purpose of resetting the background correctly when the filter gets removed again.

But I do agree that we should preserve the original background color here (e8b068d).

src/lib/filter/compile_expression.ts
GODrums: agreed, changed in 4eb72e3

Step7750 (approved):
If you get the chance, please open a PR to run the tests as a GitHub workflow.

Thanks for adding them!

PR #405, opened 2026-06-29
Fix Float Rank hyperlinks for Souvenirs/StatTraks

rxl211:
I noticed that on Steam inventories, when you selected a StatTrak or Souvenir item and it had a displayed float Rank, clicking on the link would always take you to the database with a category=1 (Normal) filter selected, even though the selected item was StatTrak/souvenir.

Example:
image

The root cause seemed to be that the itemInfo.full_item_name field was never populated, and so downstream code which looks at full_item_name to derive the correct category filter when constructing the hyperlink would always just default to category=1.

This PR includes a change to selected_item_info.ts to populate full_item_name inside of processSelectChange().

I verified this fixes the bug locally by loading the unpackaged extension in Chrome. I did not test on Firefox.

---

> [!NOTE]
> **Low Risk**
> Small change to URL query parameters for rank links; logic mirrors existing inspect-based StatTrak/Souvenir detection with no auth or data migration impact.
> 
> **Overview**
> Fixes **Rank** links on csfloat.com/db always using `category=1` (Normal) for StatTrak and Souvenir items when float rank is shown from inspect data.
> 
> `ItemInfo` now includes **`is_stattrak`** and **`is_souvenir`**, set in `fetch_inspect_info` from the decoded inspect payload (same rules already used for rank eligibility: kill-eater score type and souvenir quality `12`). **`getFloatDbLink`** in `skin.ts` picks category `2` / `3` / `1` from those flags instead of substring checks on **`full_item_name`**, which was often unset so links defaulted to Normal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec3168e7e7754bd1570433a365c363026b251755. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/lib/components/inventory/selected_item_info.ts
```diff
@@ -265,6 +265,7 @@ export class SelectedItemInfo extends FloatElement {
                     asset_id: this.asset.assetid,
                     link: this.inspectLink,
                 });
+                this.itemInfo.full_item_name = this.asset.description.market_hash_name;
```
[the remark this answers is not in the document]
rxl211: Addressed in commit
[28c14a9](https://github.com/csfloat/extension/commit/28c14a9edc61b97a0874835ee9dff95ec1e49fba) with help of Codex

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -261,12 +261,27 @@ export class SelectedItemInfo extends FloatElement {
...
                 });
+
+                if (this.asset?.assetid !== asset.assetid) {
+                    return;
+                }
+
+                this.itemInfo = {
+                    ...fetched,
+                    full_item_name: asset.description.market_hash_name,
+                };
```
GODrums: Combining data points from different sources will make this process harder to reproduce in other places (e.g. the same issue occurs in our SCM enhancements).

I would prefer reusing our existing solution from `fetch_inspect_info.ts` 
https://github.com/csfloat/extension/blob/d46eac3cd1b492fc4e33444e48ce14e448112000/src/lib/bridge/handlers/fetch_inspect_info.ts#L134-L135

and just modify the string parsing in `getFloatDbCategory` to use the available info. 

We could remove `full_item_name` from `ItemInfo` to make this a bit more clear.

GODrums (changes requested):
Thanks for the contribution and finding this bug! Added a small comment to make our parsing a bit more reliable and resolve the issue in SCM as well

src/lib/components/inventory/selected_item_info.ts
rxl211: @GODrums thanks, I wasn't thinking about SCM! I did initially look at `getFloatDbCategory` but the problem I ran into is that `ItemInfo` doesn't contain anything to infer StatTrak from. I see now that you can infer souvenir by quality == 12 though, so that can be done easily.

Are you suggesting adding an is_stattrak property to `ItemInfo` in  `fetch_inspect_info.ts`? I'm still unsure of what a common area would be to fill in the property. I tried doing so in `fetch_inspect_info.ts` but my new fields never showed up when I tried outputting the `ItemInfo` object to console.

I'm also not sure what kind of caching exists in the extension. But that's not the reason I don't think it was working for me because I tried going to SCM pages I had never been to before.

GODrums: As shown in the code sample above, you can determine if an item is StatTrak by checking if `killeaterscoretype` is defined on the decoded inspect link.

Given that these checks are simple, there should be no need to persist them in `itemInfo`, you may just replace the string parsing with those checks in `getFloatDbCategory`

rxl211: I got it working with the decoded inspect link, but since `getFloatDbCategory` didn't seem to already have an inspect link handy I had to plumb it through.

GODrums: My bad, missed that. In that case, feel free to annotate those states (stattrak, souvenir) in `ItemInfo`.
Performance-wise we should avoid running multiple decodings if not absolutely necessary, especially not in injected scripts.

rxl211: @GODrums ok thanks. I have a version working where the fields are set in `processInspectItem()` but I initially wasn't seeing my new fields show up in the ItemInfo that `getFloatDbCategory` had. At first I fixed it by deleting `threshold_cache` in Chrome devtools but later wasn't able to repro.

In short, are there any cache considerations to keep in mind when existing users upgrade / are there cases where ItemInfo does not update (and thus skips the new fields) for items users have recently looked at or something like that?

Either way, I reverted the last change and pushed the latest version so you can take a look. I'm going out of town for 7 days so won't be able to get to any edits until next week.

src/lib/utils/skin.ts
```diff
@@ -74,13 +74,14 @@ enum OrderType {
  */
 function getFloatDbLink(info: ItemInfo, order: OrderType): string {
     function getFloatDbCategory(item: ItemInfo): number {
-        if (item.full_item_name?.includes('StatTrak')) {
+        if (item.is_souvenir == null || item.is_stattrak == null) {
+            return 0; //unfiltered
```
GODrums: Is this check necessary? Applying the "normal" category should work for non-skins as well afaik

src/lib/components/inventory/selected_item_info.ts
GODrums: Thanks, looking much more simple now!

Most likely this happened because the extension's HMR in dev-mode only reloads changed files (but not cache-related files for example). This is generally not a concern for production extension updates as those reload the whole extension.

GODrums (changes requested):

GODrums:
Please also ensure to run the linter so our tests pass (`npm run lint`)!

src/lib/utils/skin.ts
rxl211: My thinking here was that under normal circumstances skins should have both these fields set, so this is meant to add some resiliency against some unexpected upstream change that occurs in which case we just show an unfiltered view of the DB instead of defaulting to Normal.

rxl211:
I think the tests should pass now. Looked like Prettier was failing on a stray blank-line whitespace. That passes now on my machine.

npm run checkformat in my Windows workspace is reporting many unchanged files because of line end differences but I don't think that'll show up when Github runs the tests.

src/lib/utils/skin.ts
GODrums: In those cases we would probably rather hard fail (i.e. defaulting to normal). Secondary issue with this check is that `==` is a type-coercion check, which you should generally never do as it comes with unexpected side effects.

Let's just remove this and this would be ready-to-merge

rxl211: done

GODrums (approved):

PR #410, opened 2026-07-28
feat(inventory): SkinCraft 3D viewer embed

ykns3:
## Summary

Adds a **View in 3D** action to Steam inventory items, opening the SkinCraft viewer in a modal over the page. The viewer is a persistent, sandboxed cross-origin iframe driven entirely over `postMessage` — the extension never re-`src`s it, so switching items inside the modal is instant once warm.

## How it works

- **`gSkinCraftEmbed`** is the single source of truth for the boot/load lifecycle: a versioned protocol (origin-pinned in both directions, never `'*'`), load correlation ids so stale `progress`/`loaded`/`error` events can't clobber a newer load, a 20s load timeout, and `pause`/`resume` on tab visibility.
- **Two worlds.** The page-world instance reads `g_ActiveInventory` and posts the clicked item plus a bounded inventory snapshot to the content-script instance, which owns the iframe. The iframe deliberately stays out of the page world, so Steam's own scripts can't reach it, its message channel, or its closed shadow root.
- **Item identity** is the masked inspect hex from `asset_properties[propertyid=6]` — SkinCraft resolves it client-side, so there's no Steam GC round-trip and no dependency on a float fetch.
- **`SkinCraftViewerModal`** owns one `<dialog>` created once and never unmounted, since unmounting would reload the embed. It renders with `lit-html`'s `render()` rather than extending `FloatElement`: the content script builds it, and Chrome leaves `customElements` null in isolated worlds, so no custom element can be defined there.
- **Inventory strip** inside the modal for moving between loaded skins without closing it, with seed/float badges from the existing float cache and card colours from Steam's own rarity/background tags. Wrapped in `guard()` so load-progress ticks don't re-diff every card.
- **WebGPU gating** (`webgpu_availability.ts`): adapter-only probe, so the button renders only where the viewer can actually run. Three-state and pessimistic while probing, so it never appears and then fails.
- **Keyboard focus** is handed to the iframe on open and on every inventory-strip selection, since the embed's hotkeys listen inside its own document.

## Sandbox / security

`sandbox="allow-scripts allow-same-origin allow-downloads"`, `allow="fullscreen"`, `referrerpolicy="no-referrer"`. Inbound messages are gated on `event.origin` and shape-validated at both boundaries (`skincraft_embed_protocol.ts`, `skincraft_viewer_protocol.ts`); outbound is pinned to the embed origin. The inventory snapshot is length-bounded and every field is checked before it crosses.

## Behaviour notes

- The 3D button also appears on other users' inventories — viewing is deliberately independent of listing eligibility, unlike *List on CSFloat*.
- *List on CSFloat* now sits in a flex row alongside the new button. That's the only part of this PR that changes existing UI rather than adding to it.
- Clients without a WebGPU adapter see no button at all. `navigator.gpu` isn't Baseline — absent before Firefox 141, platform-dependent after — against a `strict_min_version` of 127.

## Dev notes

- Local dev points `skincraft_embed_origin` at a local SkinCraft dev server. The prod embed allowlist has no localhost entries, and `steamcommunity.com` still needs adding as an allowed ancestor before this works outside dev — that SkinCraft deploy ships alongside this. Without either, the modal shows its error state; nothing else is affected.
- If the dev origin serves a self-signed cert, visit it top-level once and accept it first: Chrome refuses an untrusted-cert iframe silently, with no interstitial.

## Rollout & testing

- Local runs of the CI gates are green: `npm run build`, `npm run checkformat`, `npm test` (30 tests / 5 files), plus `tsc --noEmit` and `madge --circular` (345 files, no cycles). Node 24, matching CI.
- Tests cover the two protocol validators, progress clamping/monotonicity, and inventory extraction. The modal is verified by hand.
- Merges cleanly into current master (branch base is `master@f882f20`, unchanged).

---

> [!NOTE]
> **Medium Risk**
> New cross-origin iframe and dual postMessage boundaries are security-sensitive but origin-validated and sandboxed; inventory UI changes are localized with protocol tests covering validators.
> 
> **Overview**
> Adds a **View in 3D** flow on Steam inventory item panels that opens SkinCraft in a modal with a **single persistent sandboxed iframe** (no re-`src` on item switches). **`gSkinCraftEmbed`** owns the lifecycle: page world posts a validated open message + bounded inventory snapshot to the content script, which drives the embed via a versioned, origin-pinned `postMessage` protocol (load ids, timeouts, pause/resume on tab hide).
> 
> **`SkinCraftViewerModal`** is lit-html (not a custom element) with a closed shadow root, loading/error UI, and an optional inventory strip (seed/float from float cache, rarity styling). Items are keyed off Steam `asset_properties` inspect hex via new **`skincraft_inventory_targets`** helpers; **`float_fetcher.getCached`** exposes cache for strip badges.
> 
> **WebGPU** is probed once (`webgpu_availability` + browser-specific **`webgpu_guidance`**); the button stays hidden while checking, shows disabled + tooltip when unavailable, and opens the viewer when an adapter is available. **`skincraft_embed_origin`** is added to all environment files. **List on CSFloat** is refactored into a flex row beside the new button, with the list modal rendered outside the row to avoid layout shift.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f8fcbc5585e39c940204d964239879f21d8a69c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/lib/components/common/skincraft_viewer_modal.ts
```diff
@@ -0,0 +1,381 @@
...
+                            class="${revealed ? 'loaded' : ''}"
+                            src="${this.embedSrc}"
+                            title="SkinCraft 3D viewer"
+                            referrerpolicy="no-referrer"
+                            sandbox="allow-scripts allow-same-origin allow-downloads"
+                            allow="fullscreen"
+                        ></iframe>
+                        <div class="loading-cover ${revealed ? 'loaded' : ''}">
+                            ${this.renderLoading()}${this.renderError()}
+                        </div>
```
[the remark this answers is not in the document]
ykns3: Good catch — fixed in e19a47d. `revealed` now only covers the `revealed` phase, so an error brings the loading cover back with the retry UI interactive.

src/lib/components/common/skincraft_viewer_modal_styles.ts
```diff
@@ -0,0 +1,487 @@
...
+export const MODAL_TRANSITION_MS = 200;
+
+export const skinCraftViewerModalStyles = `
+    :host {
+        color: #f5f8ff;
+        font-family: Arial, Helvetica, sans-serif;
+    }
+
+    dialog {
+        width: min(1120px, calc(100vw - 48px));
```
Step7750, line 11: IMO, we can make this modal quite a bit larger.

On 1440p, it looks like so:

Image

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -151,8 +204,14 @@ export class SelectedItemInfo extends FloatElement {
             );
         }
 
+        if (this.canListOnCSFloat || this.canViewInSkinCraft) {
+            containerChildren.push(
+                html`<div class="market-btn-row">${this.renderListOnCSFloat()} ${this.renderViewIn3D()}</div>`
+            );
```
GODrums, line 228: This button moves a little when clicking the List button

https://github.com/user-attachments/assets/34136f44-91e0-4dd4-9aea-af466038db70

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -64,6 +77,43 @@ export class SelectedItemInfo extends FloatElement {
...
+            }
+
+            .view-3d-btn:hover {
+                filter: brightness(1.1);
+            }
+
+            .view-3d-btn:active {
+                transform: scale(0.98);
+                filter: brightness(0.95);
+            }
```
GODrums: you can nest these styles to make it easier to read, i.e.:
```css
.view-3d-btn {
  ...

  img { ... }

  &:hover {}
  &:active {}
}
```

src/lib/components/inventory/skincraft_viewer_modal.ts
```diff
@@ -0,0 +1,395 @@
...
+                aria-labelledby="skincraft-viewer-title"
+                @cancel="${this.handleCancel}"
+                @pointerdown="${this.handleDialogPointerDown}"
+                @click="${this.handleDialogClick}"
+                @transitionend="${this.handleTransitionEnd}"
+            >
+                <header class="modal-header">
+                    <div class="modal-title" id="skincraft-viewer-title">
+                        <span>${this.target?.name ?? ''}</span>
+                        <span class="modal-brand"> — SkinCraft 3D Viewer</span>
```
GODrums: Using the logo would make this look a lot more appealing. I like the way we implemented this on csfloat.com, maybe we keep it similarly here?

src/lib/components/common/skincraft_viewer_modal.ts
```diff
@@ -0,0 +1,395 @@
+import {html, nothing, render} from 'lit';
```
GODrums, line 1: we will likely want to reuse this modal for the market as well in the future so I'd move it into the `/common` folder and see if we can keep the inputs as generic as possible

src/lib/services/skincraft_viewer_protocol.ts
```diff
@@ -0,0 +1,66 @@
...
+        (target.float === undefined || (typeof target.float === 'string' && target.float.length <= 64)) &&
+        (target.rarityColor === undefined ||
+            (typeof target.rarityColor === 'string' && HEX_COLOR_PATTERN.test(target.rarityColor))) &&
+        (target.backgroundColor === undefined ||
+            (typeof target.backgroundColor === 'string' && HEX_COLOR_PATTERN.test(target.backgroundColor))) &&
+        (target.iconUrl === undefined ||
+            (typeof target.iconUrl === 'string' &&
+                target.iconUrl.length <= 4096 &&
+                target.iconUrl.startsWith(STEAM_ECONOMY_IMAGE_PREFIX)))
+    );
```
GODrums, line 61: We should distinguish between type checking the message data and validating its content. Mixing these in such a way can get very confusing

src/lib/services/webgpu_availability.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+/**
+ * Browser-level WebGPU capability check, gating every 3D affordance.
+ *
+ * `navigator.gpu` is not Baseline — absent in Firefox before 141 and platform-dependent after
+ * (https://developer.mozilla.org/en-US/docs/Web/API/Navigator/gpu#browser_compatibility), and the
+ * manifest supports Firefox 127+. Those clients see no 3D button at all, which is useful context
+ * for support tickets asking where it went.
```
GODrums: Should we display a disabled button with an informative tooltip instead of just not showing the button at all? I remember we had some user feedback for the website integration regarding this behavior.

Furthermore, this decision is made from the caller, so the comment should probably move to where that decision is made (there might be more callers in the future using this service differently).

GODrums (changes requested):
Looking forward to using this myself 👀

src/lib/components/inventory/selected_item_info.ts
ykns3: Fixed in 253001d — the list modal was mounting inside the `.market-btn-row` flex container, so its host picked up the row's 10px gap. It renders as a sibling of the row now.

src/lib/components/inventory/selected_item_info.ts
ykns3: Done in 16f22b4. Note bare-element nesting needs Chrome 120+ against our floor of 109 — older builds just drop the nested rules (icon tint, hover/active), which seemed acceptable.

src/lib/components/inventory/skincraft_viewer_modal.ts
ykns3: Done in a7df2c5 — same attribution as csfloat.com: logo mark + skincraft.gg wordmark linking to the item's SkinCraft page. The mark is inlined so it still renders when the embed origin is unreachable.

src/lib/components/common/skincraft_viewer_modal.ts
ykns3: Done in a7df2c5. Modal lives in `components/common/` with an options-object constructor and generic naming (`setItems()`, `.item-panel`/`.item-card`); the only inventory-specific bit left is the call site passing `itemsTitle: 'Inventory'`.

src/lib/services/skincraft_viewer_protocol.ts
ykns3: Split in 74f8461: `isSkinCraftItemShape` does the structural checks, `isValidSkinCraftItem` the content rules, `isSkinCraftItem` composes them. Behavior unchanged, tests pass as-is.

src/lib/services/webgpu_availability.ts
ykns3: Done in 3362c60 — muted `aria-disabled` button with a tooltip explaining why and how to recover, reusing phoenix #1687's reason split (`no-webgpu` vs `no-adapter`) and copy so both surfaces tell one story. The visibility comment moved to the caller as suggested.

src/lib/services/skincraft_inventory_targets.ts
```diff
@@ -0,0 +1,115 @@
...
+              activeInventory,
+          ]
+        : [activeInventory];
+
+    for (const inventory of inventories) {
+        const properties = inventory?.m_rgAssetProperties?.[assetId];
+        if (properties?.length) return properties;
+    }
+    return undefined;
+}
```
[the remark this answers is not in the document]
ykns3: Good catch — fixed in 20c1c96. `getLoadedInventoryTargets` now reuses `getActiveInventoryAssetProperties` per asset, so the strip consults the appwide parent map exactly like the clicked-item path; added a test for the parent-map layout.

src/lib/services/skincraft_embed.ts
```diff
@@ -0,0 +1,272 @@
...
+            this.modal?.setLoading(null);
+        } else {
+            this.modal?.showFrame();
+        }
+        this.startLoadTimeout();
+
+        if (this.ready) {
+            this.pendingInspect = undefined;
+            this.sendLoad(inspect);
+        }
```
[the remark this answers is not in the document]
ykns3: Good catch — fixed in 7ae9bc2. A load timeout while the embed never became `ready` now marks the frame for reload, and the next `requestLoad` (Retry or reopening) re-`src`s the iframe for a fresh boot under the same watchdog. A late `ready` clears the flag so a healthy embed is never reloaded.

src/lib/components/common/skincraft_logo.ts
```diff
@@ -0,0 +1,63 @@
+import {html} from 'lit';
+import type {TemplateResult} from 'lit';
+
+/**
+ * SkinCraft logo mark, mirroring csfloat.com's `assets/skincraft-logo-mark.svg`. Inlined so it
+ * renders with no network fetch — it must survive the embed origin being unreachable.
+ */
+export const skinCraftLogoMark: TemplateResult = html`
```
GODrums: We already load the image from csfloat.com for the "View in 3D" button, hence this will be cached anyways. I would just continue using the hosted version for consistency at this point.

src/lib/utils/webgpu_guidance.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+/**
+ * Reason + browser family → one-paragraph recovery copy, distilled from the csfloat.com cover so
+ * both surfaces tell one story. Internal pages (`chrome://…`, `about:…`) are spelled out inline —
+ * this renders in a CSS tooltip, which can hold neither links nor copyable text.
+ */
+export function webGpuGuidance(reason: WebGpuUnavailableReason, b: DetectedBrowser = detectBrowser()): string {
+    // Chromium forks serve their internal pages under their own scheme, not chrome://.
+    const scheme = b.isEdge ? 'edge' : b.browser === 'Opera' ? 'opera' : 'chrome';
```
GODrums: AFAIK all chromium forks also support the chrome:// command scheme. Confirmed in my local Opera version

src/lib/utils/webgpu_guidance.ts
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    switch (b.family) {
+        case 'firefox':
+            return "Firefox has WebGPU but couldn't reach your GPU. Re-enable hardware acceleration under Performance in about:preferences, then restart Firefox.";
+        case 'chromium':
+            return `${b.browser} has WebGPU but couldn't reach your GPU — graphics acceleration is usually switched off. Turn on "Use graphics acceleration when available" in ${settingsPath}, then relaunch.`;
+        default:
+            return "This browser has WebGPU but couldn't reach your GPU. Update your system and graphics drivers, then restart the browser.";
```
GODrums: "This browser supports WebGPU ... "

might sound a bit more helpful to the non-technical user

GODrums (approved):

src/lib/components/common/skincraft_logo.ts
ykns3: Done — using the hosted `csfloat.com/assets/skincraft-logo-mark.svg` and dropped the inlined copy.

src/lib/utils/webgpu_guidance.ts
ykns3: Done — one `chrome://` path for every Chromium now. That left `isEdge` unread so it's gone from `DetectedBrowser` too; Edge is still detected for the browser name in the copy.

src/lib/utils/webgpu_guidance.ts
ykns3: Done — "has WebGPU" → "supports WebGPU". Applied it to the Firefox and named-Chromium variants too so the no-adapter copy reads the same throughout.

src/lib/components/common/skincraft_viewer_modal.ts
```diff
@@ -0,0 +1,425 @@
...
+                    <div class="viewer-stage">
+                        <iframe
+                            ${ref(this.frameRef)}
+                            class="${revealed ? 'loaded' : ''}"
+                            src="${this.options.embedSrc}"
+                            title="SkinCraft 3D viewer"
+                            referrerpolicy="no-referrer"
+                            sandbox="allow-scripts allow-same-origin allow-downloads"
+                            allow="fullscreen"
+                        ></iframe>
```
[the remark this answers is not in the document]
ykns3: Not reproducing — built for Firefox in prod mode (pointing at skincraft.gg), loaded as a temporary add-on, and the modal opens and renders the viewer on a Steam inventory page. Steam's `frame-src` does exclude skincraft.gg, but Firefox isn't applying it to the content-script-injected frame, so there's nothing to relax here.

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -151,8 +219,17 @@ export class SelectedItemInfo extends FloatElement {
             );
         }
 
+        if (this.canListOnCSFloat || this.show3dButton) {
```
Step7750, line 222: We should have more of a gap between the float/paint seed and the action buttons:

Image

src/lib/components/inventory/selected_item_info.ts
```diff
@@ -151,8 +219,17 @@ export class SelectedItemInfo extends FloatElement {
...
+            // nudge the 3D button whenever it mounts.
+            containerChildren.push(
+                html`<div class="market-btn-row">${this.renderListOnCSFloat()} ${this.renderViewIn3D()}</div>
+                    ${this.renderListModal()}`
+            );
+        }
+
         if (isSellableOnCSFloat(this.asset.description)) {
-            containerChildren.push(html`${this.renderListOnCSFloat()} ${this.renderFloatMarketListing()}`);
+            containerChildren.push(this.renderFloatMarketListing());
```
Step7750, line 232: Because these are added as separate children to the container, it causes new-line overflows when it doesn't need to be: 

Image

Step7750, line 232: We should keep the CSFloat Market action button in a similar spot regardless of whether the item is listed or not.

Right now, the button is on the left if not listed and on the right otherwise.

Step7750 (commented):
Apologies for the late review! Few UI/UX comments here, perhaps suitable for a follow up.

PR #412, opened 2026-08-04
fix(inventory): scale 3D modal with viewport, unify market action row

ykns3:
Follow-up to #410, addressing @Step7750's review comments.

**Modal size** — the dialog now scales with the viewport instead of pinning to 1120px, using the same treatment as the market site's screenshot dialog: width follows 90vw, with a height-derived cap that keeps the 16:9 stage at or under 80vh. The side-panel layout (≥1520px) adds its 320px panel to the width budget; the item-strip layout (1168–1519px) adds its 122px strip to the height budget, with a 1120px floor so mid-size screens keep today's size. On a 1440p display the viewer stage grows from 1120×630 to ~1984×1116.

**Action row** — the CSFloat market action (list button or active listing, mutually exclusive) and the 3D button now share one flex row: the market action keeps the left slot whether the item is listed or not, and the buttons no longer wrap to new lines when there's room. Also adds a bit more gap between the float/paint seed info and the row.

## Validation

`npm run lint`, `npm run build`, `npm run checkformat`, `npm test` (31/31).

3d-modal-scaled

---

> [!NOTE]
> **Low Risk**
> CSS layout and inventory UI composition only; no auth, API, or data-path changes.
> 
> **Overview**
> The SkinCraft 3D viewer dialog **scales with the viewport** instead of staying near a fixed 1120px width: default sizing uses `90vw` and a **16:9 cap at 80vh** (with `dvh` fallbacks and a 1920px stage ceiling). Breakpoints for the item strip and side panel **fold those chrome dimensions into the width/height budget**, use a **1120px floor** on mid-width layouts, and drop the hard-coded stage width so the grid can grow with the dialog.
> 
> In **selected item info**, the CSFloat **list**, **active listing**, and **View in 3D** actions share **one flex row** (listing is no longer rendered below), with `stallListing` included in when that row appears so listed-only items still show the row. Spacing above the action row is slightly increased.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65e35c4e1b80e91a569b4bee3a87b6b4e0a93e89. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/lib/components/common/skincraft_viewer_modal_styles.ts
+++ b/src/lib/components/common/skincraft_viewer_modal_styles.ts
@@ -7,8 +7,11 @@ export const skinCraftViewerModalStyles = `
         font-family: Arial, Helvetica, sans-serif;
     }
 
+    /* Width follows the viewport; the height-derived term caps the 16:9 stage at 80vh so the
+       composition scales with the screen, mirroring the market site's screenshot dialog. */
     dialog {
-        width: min(1120px, calc(100vw - 48px));
+        width: min(90vw, calc(80vh * 16 / 9));
+        width: min(90vw, calc(80dvh * 16 / 9));
         max-width: none;
         padding: 0;
         border: 1px solid rgba(193, 206, 255, 0.12);
@@ -429,8 +432,10 @@ export const skinCraftViewerModalStyles = `
     }
 
     @media (min-width: 1168px) and (max-width: 1519px) and (min-height: 840px) {
+        /* The 122px item strip sits above the stage, so it joins the height budget. */
         dialog.has-items {
-            width: 1120px;
+            width: max(1120px, min(90vw, calc((80vh - 122px) * 16 / 9)));
+            width: max(1120px, min(90vw, calc((80dvh - 122px) * 16 / 9)));
         }
 
         dialog.has-items .modal-body {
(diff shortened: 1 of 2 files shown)


src/lib/components/common/skincraft_viewer_modal_styles.ts
```diff
@@ -7,8 +7,11 @@ export const skinCraftViewerModalStyles = `
         font-family: Arial, Helvetica, sans-serif;
     }
 
+    /* Width follows the viewport; the height-derived term caps the 16:9 stage at 80vh so the
+       composition scales with the screen, mirroring the market site's screenshot dialog. */
     dialog {
-        width: min(1120px, calc(100vw - 48px));
+        width: min(90vw, calc(80vh * 16 / 9));
```
Step7750: nit: You may want to set a maximum pixel width for super large screens (4k+ with odd scaling).

Step7750 (approved):
Thanks @ykns3 !

src/lib/components/common/skincraft_viewer_modal_styles.ts
ykns3: Capped the stage at 1920px (2240px with the item panel); the panel height tracks the same ceiling so it can't outgrow the stage. Only binds above ~2133x1350 CSS px, so normal displays are unchanged.

